PR #8, opened 2018-07-26
Reduce vendor.bundle.js file

dariadomagala:

- [ ] I have tested my changes.
- [ ] I have updated the relevant documentation.
---

**Description**

- Reduce vendor.bundle.js size

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "ng": "ng",
     "start": "ng serve --disable-host-check --host console-dev.kyma.local",
-    "build": "ng build",
+    "build": "ng build --prod",
     "test": "ng test",
     "lint": "ng lint",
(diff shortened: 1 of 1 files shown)


dariadomagala:
@pkosiec yesterday it worked the minute you left my desk :) I used wrong command. 
In the picture below you can see different vendor.bundle.js sizes, depending on which command was used.
screen shot 2018-07-26 at 11 38 27
Using the prod flag, the vendor chunk will be disabled by default. To show you that the vendor.bundle.js size was actually reduced, I used -vc=true which created separate vendor.bundle.js. But we don't use that option, because: 
> Total bundle sizes with Build Optimizer are smaller if there is no separate vendor chunk because having vendor code in the same chunk as app code makes it possible for Uglify to remove more unused code.

PR #11, opened 2018-07-30
Error handling for login

y-kkamil:


core/src/app/content/login-error/login-error.component.ts
```diff
@@ -0,0 +1,33 @@
...
+  styleUrls: ['./login-error.component.scss']
+})
+export class LoginErrorComponent implements OnInit {
+  constructor(private router: Router) {}
+
+  private errorMsg = '';
+  private possiblyMissingCert = false;
+
+  ngOnInit() {
+    const loginError = sessionStorage.getItem('loginError');
```
y-kkamil: we might want to introduce some 'constants service' or something similar to avoid magic strings/numbers cluttering our codebase. tdb.

PR #9, opened 2018-07-27
Few more LambdaUI fixes

rakesh-garimella:
Confirm these statements before you submit your pull request:

- [x] I have read and submitted the required [Contributor Licence Agreements](https://github.com/kyma-project/community/blob/master/CONTRIBUTING.md#agreements-and-licenses).
- [x] This pull request follows the contributing guidelines, recommended Git workflow, and templates.
- [x] I have tested my changes.
- [ ] I have updated the relevant documentation.
---

**While creating a new Lambda unless lambda name is provided user cannot interact with other components on lambda page**

Changes proposed in this pull request:

- Now unless the lambda name is not entered
   user cannot interact with other components

PR #15, opened 2018-07-30
If the lambda form is not valid we should not offer user to Save

rakesh-garimella:

---

**If the lambda form is not valid we should not offer user to Save**

 - If there is some error in the lambda form then dont allow user to save the lambda  changes
 - Improve the regex for validating the Environment Variables


rakesh-garimella:
The changes in this PR are being taken care in this[PR](https://github.com/kyma-project/console/pull/9)

PR #16, opened 2018-07-30
Handle login errors

y-kkamil:


core/src/app/content/login-error/login-error.component.ts
```diff
@@ -0,0 +1,32 @@
...
+  public possiblyMissingCert = false;
+
+  ngOnInit() {
+    const loginError = sessionStorage.getItem('loginError');
+    if (loginError) {
+      const parsedError = JSON.parse(loginError);
+      this.errorMsg = parsedError.type;
+
+      if (
+        window.location.protocol === 'http:' &&
```
kwiatekus: would not work for console served via https, i.e 'https://console.kyma.local/'

PR #20, opened 2018-07-31
Enforce https for microfrontends

y-kkamil:

- [X] I have read and submitted the required [Contributor Licence Agreements](https://github.com/kyma-project/community/blob/master/CONTRIBUTING.md#agreements-and-licenses).
- [X] This pull request follows the contributing guidelines, recommended Git workflow, and templates.
- [X] I have tested my changes.
---

### DIFF
--- a/core/src/app/extensibility/services/extensions.service.ts
+++ b/core/src/app/extensibility/services/extensions.service.ts
@@ -24,7 +24,7 @@ export class ExtensionsService {
   }
 
   isUsingSecureProtocol(url: string) {
-    if (url.substr(0, 5) !== 'https') {
+    if (url.startsWith('https')) {
       console.error(
         `${url} is not using secure protocol. External views have to be served over HTTPS.`
       );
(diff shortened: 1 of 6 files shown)


core/src/app/extensibility/services/extensions.service.ts
```diff
@@ -22,4 +22,14 @@ export class ExtensionsService {
       });
     });
   }
+
+  isUsingSecureProtocol(url: string) {
+    if (url.substr(0, 5) !== 'https') {
```
dariadomagala: You could use url.startsWith('https://'). In my opinion it's just more readable.

core/src/app/extensibility/services/extensions.service.ts
```diff
@@ -22,4 +22,14 @@ export class ExtensionsService {
...
+      console.error(
```
dariadomagala: We should also show a message in the UI, not only in dev tools.

PR #52, opened 2018-08-22
Improve labels in lambda ui

sayanh:
**Description**

- Fix modifications after creation of the labels 
- Fix modification of a value for an existing label
- Fix presentation of labels in uppercase 
- Removed `Created-by:kubeless` label
- Fixed validation of labels
- Allow labels with (starting and ending with)spaces in key or value or both 
- Allow addition of more than 3 labels

- [ ] I have updated the relevant documentation.
---

### DIFF
--- a/lambda/src/app/lambdas/lambda-details/lambda-details.component.html
+++ b/lambda/src/app/lambdas/lambda-details/lambda-details.component.html
@@ -35,7 +35,7 @@
           <div class="tn-form__item">
             <label class="tn-form__label" for="input-2">Labels</label>
             <div class="labels-input" id="input-2">
-              <span class="tn-badge tn-badge--pill labels" *ngFor="let label of labels" (click)="updateLabel(label)">
+              <span class="tn-badge-labels tn-badge tn-badge--pill labels" *ngFor="let label of labels" (click)="updateLabel(label)">
                 {{ label }}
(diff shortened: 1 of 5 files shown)


rakesh-garimella:
I have following issues here:
1. `blur` has been removed. so now we have to press enter to finalize the label. Previously clicking outside the label box would finalize the label. This new behaviour affects the editing of labels, when are in edit mode for one label and enter edit mode of the second, the previous label which was in edit mode is deleted.
2. 90px label length is confusing. Because:
     * if user has a long label text he would only be able to see if he has entered it properly after the lablel is finalized
    * if the user wants to enter the label he has to click within this 90px width to be able to enter the label
3. if a user has more than one label, editing the label pushes it to the end of the label list. This can be confusing when user is editing first label.
4. Show a different message if the label is duplicate

sayanh:
@rakesh-garimella 
> blur has been removed. so now we have to press enter to finalize the label. Previously clicking outside the label box would finalize the label. This new behaviour affects the editing of labels, when are in edit mode for one label and enter edit mode of the second, the previous label which was in edit mode is deleted.
Right, `blur` is brought back again

> 90px label length is confusing.
As discussed with @joek, `width` is set to `auto`

> if a user has more than one label, editing the label pushes it to the end of the label list. This can be confusing when user is editing first label.
As discussed with @joek, lets keep it as is

> Show a different message if the label is duplicate
Done

PR #57, opened 2018-08-24
Fix for login errors

y-kkamil:

- [X] I have read and submitted the required [Contributor Licence Agreements](https://github.com/kyma-project/community/blob/master/CONTRIBUTING.md#agreements-and-licenses).
- [X] This pull request follows the contributing guidelines, recommended Git workflow, and templates.
- [X] I have tested my changes.
---

### DIFF
--- a/core/src/app/content/login-error/login-error.component.ts
+++ b/core/src/app/content/login-error/login-error.component.ts
@@ -14,19 +14,14 @@ export class LoginErrorComponent implements OnInit {
   ngOnInit() {
     const loginError = sessionStorage.getItem('loginError');
     if (loginError) {
-      try {
-        const parsedError = JSON.parse(loginError);
-        this.errorMsg = parsedError.type || parsedError.statusText;
+      const parsedError = JSON.parse(loginError);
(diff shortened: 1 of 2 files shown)


core/src/app/content/login-error/login-error.component.ts
```diff
@@ -14,19 +14,14 @@ export class LoginErrorComponent implements OnInit {
...
+      this.errorMsg = parsedError.type;
 
-        if (
-          (parsedError.type === 'discovery_document_load_error' &&
-            parsedError.reason.status === 0) ||
-          (parsedError.url === null && parsedError.status === 0)
-        ) {
-          this.possiblyMissingCert = true;
-        }
-      } catch (err) {
```
dariadomagala: Shouldn't we keep that try/catch block? Are we sure that we will always get a stringified object?

y-kkamil: the only place that sets 'loginError' is login.component that already stringifies it. imo it would not be necessary.

PR #77, opened 2018-09-03
Console angular 6

pekura:

- [x] I have updated the relevant documentation.
---

**Description**

-
-
-

**Issue link**


components/angular/generic-list/src/app/modules/list/generic-list/generic-list.component.ts
```diff
@@ -49,6 +50,9 @@ export class GenericListComponent implements OnChanges, OnInit {
 
   setLoaded(status: boolean): void {
     this.loaded = status;
+    if (status) {
```
jesusreal: IMHO, we would reflect better what do we use "initilized" flag for, and also will set it only once even if we reload, if we write this block (also present in plain-list.component) like this:
```
    if (!this.initialized && status) {
      this.initialized = status;
    }
```

PR #95, opened 2018-09-05
Improve instance provisioning process in Service Catalog.

### LINKED ISSUE #80: One step provisioning flow
Confirm these statements before you submit the issue:

- [x] I have searched open and closed issues for duplicates.
- [x] I have read the contributing guidelines.
---

**Description**

- remove the `circle` wizard steps indicators
- all shows up as step one
- `Name` stays. Improve validation:
  - When you load a provision form to the user and we already know that for example `redis-instance` name already exists, then we should generate `redis-instance-2` name and increment in case more instances are coming. 
  - disallow name with spaces or special characters
- `Plan` and `Labels` stays as it is for now. Only look if it is possible to place name, plan and labels next to each other in one row, if it will look better.
- Then below above fields have a clear separation ruler and input parameters below, the one coming from plan schema
- modal form scale from maximum from top to bottom of the screen. In case not all fits the screen show scroll bar to the user to indicate scrolling is needed
- update tests if needed
- update OSBA contract docs

**Reasons**

- simplify provisioning
- avoid making magic solutions in 2 step wizard to change it into one step wizard in case custom plan schema is provided or not

magicmatatjahu:
**Description**

- Improve provisioning process of instance in Service Catalog (one step, not two),
- Randomly name of instance,
- Change validation regex for name of instance to simple one,
- Improve `components/react`: add scrollbar in modal content,
- Bump `components/react` version to `0.0.20`,
- Update relevant integration `catalog-ui` tests,
- Update OSBA contract docs,

**Related issue(s)**
Resolves #80


catalog/src/commons/helpers.js
```diff
@@ -95,3 +95,2873 @@ export function compareTwoObjects(obj1, obj2) {
...
+(c) by Thomas Konings
+Random Name Generator for Javascript
+*/
+export function randomNameGenerator() {
+  function getRandomInt(min, max) {
+    return Math.floor(Math.random() * (max - min)) + min;
+  }
+
+  const name1 = [
+    'abandoned',
```
michal-hudy: Please create unit test that will validate if names are following our naming convention.

PR #142, opened 2018-09-14
Improve http trigger 3

venturasr:
**Description**

- Token issuer and jwks URL can be set in lambda UI 
 - Load presetes 

**Related issue(s)**


venturasr:
See PR, https://github.com/kyma-project/console/pull/144

PR #145, opened 2018-09-14
Bugfix infinite loop on 401 errors

maxmarkus:
Redirecting to request-error component on 401 if the token is younger then 10 seconds, else go to overview.

Fixes kyma/#635


core/src/app/auth/token.interceptor.ts
```diff
@@ -9,10 +9,25 @@ import { OAuthService } from 'angular-oauth2-oidc';
 import { AppConfig } from './../app.config';
 import { Observable } from 'rxjs';
 import { tap } from 'rxjs/operators';
+import { Router } from '@angular/router';
+
+interface ITokens {
+  id_token_stored_at?: string;
+  id_token_expired_at?: string;
```
jesusreal: It look like we are not using this variable :)

maxmarkus: We're not using this whole interface anymore ^^.. thanks

core/src/app/auth/token.interceptor.spec.ts
```diff
@@ -21,11 +50,20 @@ describe('TokenInterceptor', () => {
   let http: HttpClient;
 
   beforeEach(() => {
+    spyOn(sessionStorage, 'getItem').and.callFake(mockSessionStorage.getItem);
+    spyOn(sessionStorage, 'setItem').and.callFake(mockSessionStorage.setItem);
+    spyOn(sessionStorage, 'removeItem').and.callFake(
```
dariadomagala: Why do we have to set storage item and then remove it before each test?

maxmarkus: This is setting up spies and exchanges the dom implementation with custom one. Enables us to watch on getItem calls while still having the logic working.

PR #154, opened 2018-09-18
Add info that resource quota limit has been exceeded

dariadomagala:
**Description**

- Add info that resource quota limit has been exceeded


core/src/app/content/environments/services/environments.service.ts
```diff
@@ -91,4 +95,27 @@ export class EnvironmentsService {
         );
     }
   }
+
+  public getResourceQueryStatus(environment: string) {
+    const query = ` query ResourceQuotasStatus($environment: String!) {
```
y-kkamil: maybe i'm too picky here, but there's unnecessary space at the very begging of the query :P

core/src/app/content/environments/environments-container/environments-container.component.ts
```diff
@@ -127,4 +141,42 @@ export class EnvironmentsContainerComponent implements OnInit, OnDestroy {
     }
     return true;
   }
+
+  private hideError() {
```
y-kkamil: it has to be public probably, so that the template can access it
https://stackoverflow.com/questions/43141576/property-x-is-private-and-only-accessible-within-class-xyzcomponent

Issue #136, opened 2018-09-13
Make logo and title in console configurable


hardl:
... so console can be rebranded

PR #155, opened 2018-09-18
#136 Make logo and title in console configurable

hardl:
**Description**

- logo and title in console configurable via configmap


core/src/app/app.config.ts
```diff
@@ -40,6 +40,8 @@ const config = {
   lambdasModuleUrl: `https://lambdas-ui.${domain}`,
   orgId: 'my-org-123',
   orgName: 'My Organization',
+  headerTitle: '',
```
kwiatekus: Shouldn't we define the fallback values  here (rather than in the component?) We do that for organisation id and name. Empty strings look suspicious

hardl: Yes and no ;)
We could do that for logo but not for headerTitle, since its default value IS empty string.

PR #157, opened 2018-09-18
Allow ServiceBinding without secret

michal-hudy:
**Description**

- Allowed ServiceBinding without secret in UI

**Related issue(s)**

See kyma-project/kyma#372


instances/src/components/ServiceInstanceDetails/ServiceInstanceBindings/ServiceInstanceBindings.component.js
```diff
@@ -60,14 +60,16 @@ class ServiceInstanceBindings extends React.Component {
           size: 0.2,
           accesor: el => {
             const secret = el.serviceBinding && el.serviceBinding.secret;
-            return (
+            return secret ? (
```
magicmatatjahu: Change to `secret && Object.keys(secret).length ?`

Issue #133, opened 2018-09-11
The first load of catalog, content and service instances is very slow


michal-hudy:
**Description**

The first load of `catalog`, `content` and `service instances` is very slow. It is caused by big size of `main.js`.

File sizes:
 - `catalog`: 6.4MB
 - `service instances`: 6.4 MB
 - `content`: 5.8 MB

**Expected result**

Size of these files should be reduced.

mjasinski5:
https://github.com/kyma-project/console/pull/158

mjasinski5:
Improvements:
**components/react**
* Webpack configuration changed
    * Created weback.production.config
    * Devtool option set to false in production mode
    * NODE_ENV **production** passed in (DefinePlugin)
* Content
    * Improvements from **components/react** affected size heavily
    * changed webpack configuration to optimise react-hightlightjs. Now it only bundles few languages instead of all.
* Catalog and Instances
    * Improvements from react-components affected size heavily
    *  CodeSplitting - dynamic imports for swagger-ui library, so its not included in the main bundle

Other stuff:
* Ive also added npm run analyze command to the package.json’s. It will help to analyze bundles in the future.
* Some kind of test should be created to check sizes of the bundles - https://github.com/kyma-project/console/issues/161

Future improvements:
* We need to find a way to reduce size of the components/react. Currenly, for example in the catalog ui, it takes 50.4% of the final size. It applies to all views.
    * Ideas: In the **components/react** we are using react-jsonschema-form in one of the components. It costs us 12% of the final bundle size. Maybe there’s a way to avoid that (example: move it from components -> particular projects + dynamic import like I did it with swagger-ui) 
    * We are using dependencies (e.g schema-utils) which requires ajv library which is VEEERY BIG. ATM it takes 16% of the final bundle.
    * Change the way library is built. Now if we include one components, all bundle is included. Do it as for example material-ui did it.

IMHO we need interation 2 for further improvements.

Final sizes of the bundles:
* instances - 1014 KB
* Catalog - 1.1 MB
* Content - 991 KB

With maximum optimisation of **components/react** we will be able to reduce it with additional ~500kb

Cheers,
  Mariusz

mjasinski5:
Review: @michal-hudy @magicmatatjahu

PR #158, opened 2018-09-19
Reduce size of the React UI bundles

mjasinski5:
**Description**

Improvements:
* components/react
    * Created weback.production.config
    * Devtool option set to false in production mode
    * NODE_ENV **production** passed in (DefinePlugin)
* Content
    * Improvements from **components/react** affected size heavily
    * changed webpack configuration to optimise react-hightlightjs. Now it only bundles few languages instead of all.
* Catalog and Instances
    * Improvements from react-components affected size heavily
    *  CodeSplitting - dynamic imports for swagger-ui library, so its not included in the main bundle

**Related issue(s)**

#133


mjasinski5:
resolves #133

Issue #179, opened 2018-09-25
Make UI tests stable


derberg:
**Description**
Investigate the issue with running UI tests - find the reason of failing tests
- [x] make sure tests go through on local and cluster (or at least find the reason)
- [x] check the resources limit, maybe it is enough to run test in another environment then kyma-system
- [x] check if we need to enhance test logic to make it runnable on stability checks, like we did with ui-api-layer - skip them when domain log-in is enabled, like in ui-api-layer

**Expected result**

If UI test fails, it fails because some functionality was broken and UI doesn't work anymore 

**Actual result**

UI tests fail randomly for different unspecified reasons

**Steps to reproduce**

Run tests on any cluster several times

pkosiec:
Review: @pkosiec @akucharska

PR #185, opened 2018-09-26
Fix UI Acceptance tests

michal-hudy:
**Description**

- Fixed docker image
- Code cleanup
- Fixed content ui
- Fixed tests

**Related issue(s)**

See #179

### DIFF
--- a/tests/ui-tests/utils/address.js
+++ b/tests/ui-tests/utils/address.js
@@ -1,0 +1,59 @@
+import config from '../config';
+
+// Console
+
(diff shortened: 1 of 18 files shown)


tests/ui-tests/utils/address.js
```diff
@@ -0,0 +1,59 @@
+import config from '../config';
+
+// Console
```
pkosiec: Instead of comments we can define classes, instantiate them and export their instances

PR #227, opened 2018-10-12
 Display graphql error for remote environments creation and improve graphql errors handling

dariadomagala:
**Description**

- Display graphql error for remote environments creation 
- Improve graphql errors handling

**Related issue(s)**


components/angular/generic-list/src/app/modules/list/plain-list/plain-list.component.ts
```diff
@@ -109,12 +109,17 @@ export class PlainListComponent implements OnChanges {
       error => {
         this.setLoaded(true);
         this.setLoading(false);
-        this.errorMessage = error.message ? error.message : error.error;
+        this.errorMessage = error.error
```
jesusreal: If possible I would make this code more readable by using regular `if` blocks

core/src/app/shared/components/labels-input/labels-input.component.ts
```diff
@@ -35,21 +35,25 @@ export class LabelsInputComponent implements OnInit {
   }
 
   public validateNewLabel() {
-    this.setWrongLabelMessage(this.newLabel);
-    this.labelsChangeEmitter$.emit({
-      wrongLabels: Boolean(this.wrongLabelMessage)
-    });
+    if (this.newLabel) {
```
jesusreal: Is this if block really necessary? For me it looks it isn't

core/src/app/shared/components/labels-input/labels-input.component.ts
```diff
@@ -35,21 +35,25 @@ export class LabelsInputComponent implements OnInit {
...
   public addLabel() {
     this.validateNewLabel();
     if (this.newLabel && !this.wrongLabelMessage) {
-      this.labels.push(
-        this.newLabel
-          .split(':')
-          .map(s => s.trim())
-          .join(':')
-      );
+      this.labels.push(this.newLabel.split(':').join(':'));
```
jesusreal: If we don´t need to trim anymore, we also don´t need to split and join ;)

components/angular/generic-list/src/app/modules/list/plain-list/plain-list.component.ts
```diff
@@ -109,12 +109,23 @@ export class PlainListComponent implements OnChanges {
...
         this.setLoaded(true);
         this.setLoading(false);
-        this.errorMessage = error.message ? error.message : error.error;
+
+        let errorMessage;
+        if (error.error) {
+          if (error.error.message) {
+            errorMessage = error.error.message;
+          } else {
+            error = error.message || error;
```
jesusreal: Do we have to set `error` here, or rather `errorMessage`? In case the answer is `errorMessage`, it could be shortened to:
```
if (error.error && error.error.message) {
  errorMessage = error.error.message;
} else {
  error = error.message || error;
}
```
Additionally, we might not need `errorMessage` variable, we could set `this.errorMessage` directly.

core/src/app/shared/services/graphql-client-service.ts
```diff
@@ -21,17 +21,19 @@ export class GraphQLClientService {
...
               error = err.error.errors[0].message;
-              break;
-            case err.error && err.error.message:
+            } else if (err.error.message) {
               error = err.error.message;
-              break;
-            default:
-              error = err.message;
-              break;
+            } else {
```
jesusreal: You could remove some duplication by replacing these 2 else blocks with `error = error || err.message || error`, but what you did is also good ;)

components/angular/generic-list/src/app/modules/list/plain-list/plain-list.component.ts
dariadomagala: You are totally right, changed this piece of code so many times util I got lost with that 😄

Issue #156, opened 2018-09-18
Use subscriptions for Instances view


pkosiec:
AC:
- [ ] Configure Apollo Client on UI
- [ ] enable websockets on listing instances so without page refresh
  - we see new instances are added to the list
  - particular instance disappears from the list in case it is deleted
  - status is updated from PROVISIONING to READY or any other status, and the status.message too

michal-hudy:
It is not possible to connect to UI API, because Apollo Client 2 don't support HTTP headers for WebSocket connection - https://github.com/apollographql/apollo-client/issues/3967

Also we need to consider global WebSockets for `Service Binding` and `Service Binding Usage`.

derberg:
@michal-hudy check if we can downgrade to Apollo 1 or check if we can do some other solution here, some custom stuff on istio or something

pkosiec:
Review: @michal-hudy @magicmatatjahu

PR #249, opened 2018-10-19
Use subscriptions on Instances UI

pkosiec:
**Description**

- Setup Apollo GraphQL Client for subscriptions in Instances UI
- Use WebSockets for instance view

**Related issue(s)**
#156


pkosiec:
Okay, I noticed one bug:
"Filters" option is not showing, when all instances are without labels. Working on fix right now.

instances/src/components/DataProvider/DataProvider.js
```diff
@@ -0,0 +1,99 @@
...
+    subscribeToEvents: subscribeToEvents,
+  });
+};
+
+DataProvider.propTypes = {
+  children: PropTypes.func.isRequired,
+  serviceInstances: PropTypes.object,
+};
+
+export default compose(
```
magicmatatjahu: Maybe create DataProvider.component and DataProvider.container?

pkosiec: It would be really hard to split this component into two, as IMHO 100% of it is the container

PR #250, opened 2018-10-22
Change cursor in dropdowns

### LINKED ISSUE #243: wrong cursor style for dropdown entries
It seems like all dropdown entries have "text selection" cursor, because we are using an anchor tag without setting the href attribute.

They should have "cursor:pointer".

It should be probably fixed in generic list component

parostatkiem:
**Description**

- Change cursor in `generic-list-actions` component

**Related issue(s)**
Resolves #243


parostatkiem:
I'll reopen it due to technical issues

PR #303, opened 2018-11-06
Change docker images tagging strategy

kwiatekus:
**Description**

When build is triggered on master branch, the commit hash (short) is used for tagging docker images.

**Reason**
This way we can trace back the code base of any docker image that is used as kyma component and create release maintenance branches that would originate from the given commit, i.e
```
git checkout <commitSHA>
git checkout -b release-x.x.x 
```

**Related issue(s)**
Resolves https://github.com/kyma-project/console/issues/301

### DIFF
--- a/core/src/app/content/environments/environment-details/environment-details.component.html
+++ b/core/src/app/content/environments/environment-details/environment-details.component.html
@@ -7,12 +7,6 @@
         <h1 class="sf-toolbar__header">
             <span>{{ environment.getLabel() }}</span>
         </h1>
-        <div class="sf-toolbar__right">
-            <button class="tn-button tn-button--text tn-button--small sf-toolbar__item" (click)="openUploadResourceModal()"
-                *ngIf="!showSelectFileButton">
-                + Deploy new resource to the environment
-            </button>
-        </div>
     </div>
     <div class="sf-section">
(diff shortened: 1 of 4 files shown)


orchestrator.Jenkinsfile
```diff
@@ -49,7 +49,7 @@ podTemplate(label: label) {
                             checkout scm
                             // use HEAD of branch as revision, Jenkins does a merge to master commit before starting this script, which will not be available on the jobs triggered below
                             commitID = sh (
-                                script: "git rev-parse origin/${env.BRANCH_NAME}",
+                                script: "git rev-parse --short origin/${env.BRANCH_NAME}",
```
pkosiec: I've checked how it is done on `kyma` repository, and these are my findings:

```
commitID = sh (script: "git rev-parse origin/${env.BRANCH_NAME}", returnStdout: true).trim()

(...)

appVersion = commitID.substring(0,8)
```

Maybe we should also do that - or at least give a value for `--short` parameter

```
--short[=length]
Same as --verify but shortens the object name to a unique prefix with at least length characters. The minimum length is 4, the default is the effective value of the core.abbrev configuration variable (see git-config[1]).
```

PR #307, opened 2018-11-07
Fix instances UI bugs

pkosiec:
**Description**

- Fix error counter for Websocket binding events
- Truncate columns on Instances list
- Truncate text in Service Instance details tables
- Fix console warning for all modals
- Do not show "Select existing bindings" when there are no bindings
- Obtain admin password when running local UI tests

**Related issue(s)**
#306


components/react/src/components/Modal/index.js
```diff
@@ -136,8 +136,8 @@ class Modal extends React.Component {
             style={reactModalCustomModalStyles}
           >
             <ModalHeader>
+              <div>{title}</div>
```
magicmatatjahu: If you must do it in this way, please create new styled components `ModalHeaderTitle` with display: `inline-block`, because already we have that in `Bind Application` modal in instances ui:
image

instances/src/components/ServiceInstanceDetails/ServiceInstanceBindings/StatusIndicator/StatusIndicator.component.js
```diff
@@ -3,49 +3,39 @@ import React, { Fragment } from 'react';
...
         item => item.status.type === type,
       ).length;
     }
     return statusCounts;
   };
 
-  render() {
-    const { statuses } = this.state;
-    let statusType;
+  const statuses = countByType();
```
magicmatatjahu: Please change this to more transparent name, like a `numberOfStatusesByType` or something like this.

components/react/src/components/Modal/index.js
```diff
@@ -15,6 +15,7 @@ import {
   ModalFooter,
   ModalCloseButton,
   ModalInfoButton,
+  ModalHeaderTitle,
```
magicmatatjahu: image

components/react/src/components/Modal/components.js
```diff
@@ -25,6 +26,10 @@ export const ModalHeader = styled.header`
   }
 `;
 
+export const ModalHeaderWrapper = styled.div`
```
magicmatatjahu:
```js
export const ModalHeaderWrapper = styled.div`
  display: inline-block;
`;

export const ModalHeaderTitle = ModalHeaderWrapper;
```

instances/src/components/ServiceInstanceDetails/ServiceInstanceBindings/StatusIndicator/StatusIndicator.component.js
```diff
@@ -3,49 +3,39 @@ import React, { Fragment } from 'react';
...
         item => item.status.type === type,
       ).length;
     }
     return statusCounts;
   };
 
-  render() {
-    const { statuses } = this.state;
-    let statusType;
+  const statuses = countStatusesByType();
```
magicmatatjahu: Please change `statuses ` to `numberStatusesByType`.

components/react/src/components/Modal/index.js
pkosiec: tenor-207763321

pkosiec: Done

Issue #317, opened 2018-11-09
Service RE labels are wrong in the Service Catalog


mcansimsek:
**Description**

When I bind two remote environments to one environment, the services with the same name has also same remote environment label. 

![image](https://user-images.githubusercontent.com/15628751/48271435-9db73380-e43c-11e8-8798-78e0483a296d.png)

**Steps to reproduce**

1- Bind two REs, which has the service with the same name, to the same environment
2- Open catalog and check the label of the services

mjasinski5:
@mcansimsek @lszymik  Ive seen that you unregistered those 2 classes from wormhole cluster, could you please bring it back? ;) Just for testing&review purposes, it would speed up the process.

mcansimsek:
@mjasinski5 I registered those services to "stage" environment

mjasinski5:
@mcansimsek thanks! 👍

PR #318, opened 2018-11-11
Fix condition for choosing RE

mjasinski5:
**Description**
If few service classes had the same display name, connected RE was displayed in wrong way. 

- Condition for choosing connected RE has been changed

**Related issue(s)**
https://github.com/kyma-project/console/issues/317
https://github.com/kyma-project/console/issues/319

### DIFF
--- a/catalog/src/components/ServiceClassList/ServiceClassList.component.js
+++ b/catalog/src/components/ServiceClassList/ServiceClassList.component.js
@@ -122,11 +122,13 @@ class ServiceClassList extends React.Component {
     // https://github.com/apollographql/apollo-client/issues/2920
     // Possible solution: do resolver logic on component side
     items = items.map(entry => {
-      const remoteEntry = serviceClasses.find(
-        remoteEntry =>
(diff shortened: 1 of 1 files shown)


catalog/src/components/ServiceClassList/ServiceClassList.component.js
```diff
@@ -124,7 +124,6 @@ class ServiceClassList extends React.Component {
     items = items.map(entry => {
       const remoteEntry = serviceClasses.find(
```
magicmatatjahu:
```js
const remoteEntry = serviceClasses.find(remoteEntry =>{
  if (remoteEntry.name) return remoteEntry.name === entry.name;
  if (remoteEntry.externalName) return remoteEntry.externalName === entry.externalName ;
  return remoteEntry.displayName === entry.displayName;
})
```
would be better 😄

mjasinski5: Not sure if we should take displayName into consideration. We should double check but I think **name** or **externalName** are mandatory fields anyway.

Issue #337, opened 2018-11-20
If there is no prefix passed, then the servicebinding prefix is passed as `undefined`


rakesh-garimella:
**Description**

If we are not passing any prefix then servicebinding prefix is passed as `undefined` thus breaking the example.

an example is `undefined-GATEWAY_URL`

**Expected result**

If no prefix is passed we should not add any prefix.

**Actual result**

**Steps to reproduce**
Try to create a binding with no service binding prefix. Check the environment vars. It would have `undefined-` prepended to the secret.

**Troubleshooting**

**Workaround**
Bind the service to lambda via `instances menu`

PR #339, opened 2018-11-21
if there is not prefix defined then dont insert any prefix for service binding usage

rakesh-garimella:
**Description**
Insert prefix for `servicebindingusage` only if it is defined
**Related issue(s)**
Fixes: #337


sayanh:
The following workflow breaks with HTTP 409 for sbu and sb:
- Delete an existing sb from UI which has no prefix.
- Add the same sb(create a new secret) but with prefix.

PR #340, opened 2018-11-21
Add a status column to the list of bound services

venturasr:
**Description**
Add a status column to the list of bound services

**Related issue(s)**
- https://github.com/kyma-project/kyma/issues/1593

### DIFF
--- a/lambda/src/app/lambdas/lambda-details/lambda-instance-bindings/lambda-instance-bindings.component.ts
+++ b/lambda/src/app/lambdas/lambda-details/lambda-instance-bindings/lambda-instance-bindings.component.ts
@@ -15,14 +15,20 @@ import { InstanceBindingInfo } from '../../../shared/datamodel/instance-binding-
 import {
   ServiceBindingUsage,
   IServiceBindingUsageList,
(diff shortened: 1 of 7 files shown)


lambda/src/app/lambdas/lambda-details/lambda-instance-bindings/lambda-instance-bindings.component.ts
```diff
@@ -168,4 +179,21 @@ export class LambdaInstanceBindingsComponent {
     this.instanceBindingInfoList.push(instanceBindinginfo);
     this.bindingStateEmitter.emit(this.bindingState);
   }
+
+  public determineClass(instanceBinding) {
```
rakesh-garimella: we should fetch status form `lambda-instance-binding-creator.component.ts`. Also if the status of the binding is `true` then display `green Ready` (This would happen when we are using existing service binding). 

When we are creating a new service binding then it can be `pending`

venturasr: It has been updated.

PR #359, opened 2018-11-29
Handle promise rejection in the confirmation modal

dariadomagala:
**Description**

- Handle promise rejection in the confirmation modal

**Related issue(s)**


dariadomagala:
Since the 'dirty state' feature has beed implemented, the confirmation modal rejects the promise after clicking 'Cancel' and our components ignored it.

PR #360, opened 2018-11-30
Prow migration for console

### LINKED ISSUE #353: Prow pipeline for kyma/console
- Migrate CI from Jenkins into Prow job, and make prow as master CI
- Check if we need to update documentation

y-kkamil:
**Description**

Added Makefile for prow ci.

**Related issue(s)**
Resolves #353

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -7,7 +7,7 @@ ci-master: install-dependencies validate test build build-image push-image
 ci-release: install-dependencies validate test build build-image push-image
 
 install-dependencies:
-	npm ci
+	npm install -f
 
 validate:
 	npm run lint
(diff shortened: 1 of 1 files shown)


core/Makefile
```diff
@@ -0,0 +1,30 @@
...
+TAG = $(DOCKER_TAG)
+
+ci-pr: install-dependencies lint test build validate build-image push-image
+ci-master: install-dependencies lint test build validate build-image push-image
+ci-release: install-dependencies lint test build validate build-image push-image
+
+install-dependencies:
+	npm ci
+
+lint:
```
michal-hudy: Jenkins expects that `validate` target is defined: https://github.com/kyma-project/console/blob/master/core/Jenkinsfile#L57
It is also expected by `ci*` targets, see lines 5,6,7 ;)


Change it here or in `jenkinsfile`

y-kkamil: fixed

michal-hudy:
@jesusreal Branch protection will be configured by Prow, when we will migrate. Right now Prow is not production ready, so it should not block PR. Also, as you can see it didn't send status to GH.

Issue #356, opened 2018-11-28
Catalog UI to be resilient on empty binding schemas


derberg:
**Description**

* Make sure we do not display empty modal when creating credentials
* Make sure we do not display any strange empty spaces in form for binding application
* Check if we do not have the same problem with provisioning modal, if we do not have to secure there
* Check out possible solutions described here https://github.com/Azure/open-service-broker-azure/issues/634 and discussed with @zhongyi-zhang
* Decide where to apply the fix but I think the safest and best place to do that is on ui-api-layer side, so in case of empty schema we will return null because we are already resilient to null in both views.

**Expected result**

1. We know that given class returns empty schema that cannot be used for service binding creation
2. In case of empty schema we do not display modal but act like there is no schema

**Actual result**

You click on Create Credentials and they are not created but empty modal is shown.
Reason is that it returns schema:
```
{
            "$schema": "http://json-schema.org/draft-04/schema#",
            "additionalProperties": false,
            "type": "object"
}
```

but it could also be:
```
{ 
   "description": "voice mail record", 
   "type": "object", 
   "additionalProperties": false, 
   "properties": { 
    } 
  } 
} 
```

**Steps to reproduce**

1. Provision some Azure ServiceClass
2. Go to instance details and try to create credentials (ServiceBinding)
3. You get an empty modal with no values

derberg:
During the implementation please asses how much time it would take to always display on Credentials tab (when the Binded Apps tab is active) information about a number of all Credentials, not only corrupted ones. If it doesn't affect the estimation, go ahead with implementation

mjasinski5:
Review: @akucharska @mjasinski5

PR #396, opened 2018-12-09
Check exists of bindingCreateParameterSchema field

magicmatatjahu:
**Description**

- Check exists of `bindingCreateParameterSchema` field,
- Display on `Credentials` tab (when the `Bound Apps` tab is active) information about a number of all Credentials, not only corrupted ones. And the other way around too.

**Related issue(s)**
See also #356

### DIFF
--- a/instances/src/components/ServiceInstanceDetails/ServiceInstanceBindings/CreateCredentialsModal/CreateCredentialsModal.component.js
+++ b/instances/src/components/ServiceInstanceDetails/ServiceInstanceBindings/CreateCredentialsModal/CreateCredentialsModal.component.js
@@ -145,9 +145,10 @@ class CreateCredentialsModal extends React.Component {
       bindingCreateParameters: bindingCreateParameters,
     };
 
+    const bindingCreateParameterSchemaExists = bindingCreateParameterSchema && (bindingCreateParameterSchema.$ref || bindingCreateParameterSchema.properties);
+
     const content = [
       <Fragment key={serviceInstance.name}>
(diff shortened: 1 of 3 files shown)


instances/src/components/ServiceInstanceDetails/ServiceInstanceBindings/CreateCredentialsModal/CreateCredentialsModal.component.js
```diff
@@ -145,9 +145,11 @@ class CreateCredentialsModal extends React.Component {
       bindingCreateParameters: bindingCreateParameters,
     };
 
+    const bindingCreateParameterSchemaExists = bindingCreateParameterSchema && (bindingCreateParameterSchema.$ref || bindingCreateParameterSchema.properties);
```
pkosiec: Hmm, I think the same changes should be applied in Catalog UI for `instanceCreateParameterSchema`.

magicmatatjahu: This is already implemented 😸 https://github.com/kyma-project/console/blob/master/catalog/src/components/ServiceClassDetails/CreateInstanceModal/CreateInstanceModal.component.js#L210

pkosiec: ekhm 😄

PR #407, opened 2018-12-11
Hide service instances if Service Plan has bindable attribute set to false

joek:
See kyma-project/kyma#1596


k15r:
a little bug sneaked in:
create a service binding but do not create a new secret. Instead reuse an already existing secret. Save the lambda.
now edit this lambda and remove the service binding. Click the save button.

*Expected behaviour:*
It should now go back to the lambda list. 

*Actual behaviour:*
it doesnt.

k15r:
> The problem should be already fixed by santiago (merged today).

I verified it. Its fixed

lambda/src/app/lambdas/lambda-details/lambda-instance-bindings/lambda-instance-binding-creator/lambda-instance-binding-creator.component.ts
```diff
@@ -53,23 +48,25 @@ export class LambdaInstanceBindingCreatorComponent {
       this.environment = eventData.currentEnvironmentId;
       this.token = eventData.idToken;
       this.serviceInstancesService
-        .getServiceInstances(this.environment, this.token)
+        .getServiceInstances(this.environment, this.token, 'RUNNING')
```
k15r: That could be placed in a class constant

PR #465, opened 2019-01-09
Add icons for nodes and categories

jesusreal:
**Description**

- Add icons for categories and some other nodes in left navigation

**Related issue(s)**


magicmatatjahu:
I think that `sap-box` would be better for Documentation.

PR #494, opened 2019-01-15
add johannesdoberer to codeowners list

JohannesDoberer:
add johannesdoberer to codeowners list


maxmarkus:
Would you mind adding also `marynaKhromova` so we can add you both together, please. And you need to sign the CLA (click on the cla badge)

PR #497, opened 2019-01-16
Add list of backend modules to the context

dariadomagala:
**Description**

- Add list of backend modules to the context

**Related issue(s)**


tests/ui-tests/utils/testContext.js
```diff
@@ -15,11 +15,13 @@ const context = (function() {
         ignoreHTTPSErrors: true,
         headless: config.headless,
         slowMo: 80,
+        devtools: true,
```
maxmarkus: do we want to have devtools `true` by default? also when it runs in ci

jesusreal: This belongs actually to another pr already merged to master, there was a git issue in the branch, I fixed it and this change is not there anymore.

PR #503, opened 2019-01-17
Apply backend modules data to deployments view

jesusreal:
**Description**

- Show "bound apps" column in deployments table only if there is a backend module called 'servicecatalogaddons'
- In deployments view, replace the header name "Bound apps" with "Bound services" and make the name of the bounded service instances clickable, so that the user can go to the instance details.
- Write unit tests
- Introduce ts config 'paths' feature for shorter and easier to write imports.

**Related issue(s)**

### DIFF
--- a/core/src/app/content/environments/operation/deployments/deployment-header-renderer/deployment-header-renderer.component.html
+++ b/core/src/app/content/environments/operation/deployments/deployment-header-renderer/deployment-header-renderer.component.html
@@ -3,7 +3,7 @@
   <div class="col-1">Pods</div>
   <div class="col-1">Age</div>
   <div class="col-2">Images</div>
-  <div class="col-1">Bound apps</div>
+  <div class="col-1" *ngIf="showBoundServices">Bound apps</div>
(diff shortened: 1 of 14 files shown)


core/src/app/content/environments/operation/deployments/deployment-header-renderer/deployment-header-renderer.component.html
```diff
@@ -3,8 +3,8 @@
   <div class="col-1">Pods</div>
   <div class="col-1">Age</div>
   <div class="col-2">Images</div>
-  <div class="col-1">Bound apps</div>
-  <div class="col-2">Labels</div>
+  <div class="col-1" *ngIf="showBoundServices">Bound services</div>
```
kwiatekus: I'm curious why this header was renamed ?

jesusreal: It is a requirement in the task

PR #518, opened 2019-01-24
K8s Dashboard API removal

### LINKED ISSUE #508: Remove queries to k8s dashboard api
... as a temporary solution until full switch to ui-api-layer

Consider already using implemented ui-api-layer queries if applicable ("pods" is a candidate, see https://github.com/kyma-project/kyma/issues/2000)

AC:
* Kyma console fully working without k8s dashboard deployed

y-kkamil:
**Description**

- Removes calls to k8s dashboard api and instead uses api server
- For replica sets and pods not all error/warning messages are provided due to lack of them in the api server response. These are to be added when ui-api-layer is 100% ready. 

**Related issue(s)**
Resolves #508


core/src/app/content/environments/operation/services/service-details/service-details.component.ts
```diff
@@ -42,26 +41,18 @@ export class ServiceDetailsComponent implements OnInit, OnDestroy {
...
             res => {
               this.serviceDetails = res;
               this.serviceDetailsLoading = false;
-              this.fetchData(this.serviceDetailsToEditUrl).subscribe(
-                response => {
-                  this.serviceDetailsToEdit = response;
-                  this.serviceDetailsToEditLoading = false;
-                }
-              );
+              this.fetchData(this.serviceDetailsUrl).subscribe(response => {
```
dariadomagala: We don't need that anymore. We fetched it 4 lines above. This whole component can be refactored as this `serviceDetailsToEdit` etc. is not needed anymore as well.

y-kkamil: removed

core/src/app/content/environments/operation/replica-sets/replica-sets-entry-renderer/replica-sets-entry-renderer.component.html
```diff
@@ -1,21 +1,21 @@
 <div class="row sf-list__body" [ngClass]="{'sf-list__disabled': disabled === true}">
   <div class="col-2 col-lg-2 sf-list__body--primary">
-  {{ entry.objectMeta.name }}  
+  {{ entry.metadata.name }}
   </div>
-  <div class="col-1">{{ entry.pods.running }}/{{ entry.pods.desired }}</div>
-  <div class="col-1">{{ entry.objectMeta.creationTimestamp | timeAgo }}</div>
+  <div class="col-1">{{ entry.status.readyReplicas || '0' }}/{{ entry.status.replicas }}</div>
```
dariadomagala: Something is not right here, on master I get `2/3`, and here, same replica set shows `2/2`.
What came from the k8s api:
screen shot 2019-01-28 at 12 08 00

PR #520, opened 2019-01-25
Add babel-based transpilation for luigi config file

jesusreal:
**Description**

- Add babel-based transpilation for luigi config file

**Related issue(s)**


core/.babelrc
```diff
@@ -0,0 +1,11 @@
+{
+  "presets": [
+    [
+      "@babel/preset-env",
+      // {
+        // "useBuiltIns": "usage",
+        // "targets": "> 0.25%, not dead"
+      // }
```
parostatkiem: I think it shouldn't be here. Everyone can easily find it in Babel docu :)

core/package.json
```diff
@@ -4,19 +4,19 @@
   "license": "MIT",
   "scripts": {
     "ng": "ng",
-    "start": "echo `` > src/assets/config/config.js && ng serve --disable-host-check --host console-dev.kyma.local",
-    "start:kyma": "npm run copy-config && ng serve --disable-host-check --host console-dev.kyma.local",
-    "start:kyma:api": "npm run copy-config && ng serve --disable-host-check --host console-dev.kyma.local --configuration api-local",
-    "build": "ng build --prod",
+    "start": "echo `` > src/assets/config/config.js && concurrently \"npm run buildConfig:watch\" \"ng serve --disable-host-check --host console-dev.kyma.local\"",
+    "start:kyma": "npm run copy-config && concurrently \"npm run buildConfig:watch\" \"ng serve --disable-host-check --host console-dev.kyma.local\"",
+    "start:kyma:api": "npm run copy-config && concurrently \"npm run buildConfig:watch\" \"ng serve --disable-host-check --host console-dev.kyma.local --configuration api-local\"",
```
parostatkiem: 🥇 for watch-mode! I forgot about it in Angular Example :(

core/.babelrc
jesusreal: Sure, I can remove it if you want, but I find just more convenient to leave it there because I was testing it and I made sure this is what we would need to add to the code if we are not compatible enough. It´s 5-10min work the other way. so I don't mind actually

jesusreal:
Regarding your comment for changing the location of the `luigi-config` file, I actually moved it to another folder and I am compiling it from there. But I forgot to remove the luigi-config file from the old location. Thanks for noticing, I will remove it

Issue #459, opened 2019-01-09
Visibility restrictions for nodes in console


hardl:
First step for role-based UI adjustment using group information.

AC:
* users not part of "kyma-admins" can't see cluster-scoped nodes other than "General Settings" (which is to be renamed ;) )

dariadomagala:
Using the selfSubjectRulesReview, we get something like: 
```
"status": {
    "resourceRules": [
      {
        "verbs": [
          "*"
        ],
        "apiGroups": [
          "",
          "apiGroup1",
          "apiGroup2",
          "apiGroup3",
          "apiGroup4"
        ],
        "resources": [
          "*"
        ]
      },
      {
        "verbs": [
          "get",
          "list",
          "watch"
        ],
        "apiGroups": [
          "apiGroup5"
        ],
        "resources": [
          "resource5"
        ]
      },
      {
        "verbs": [
          "*"
        ],
        "apiGroups": [
          "rbac.authorization.k8s.io"
        ],
        "resources": [
          "*"
        ]
      }
    ],
    "nonResourceRules": [
      {
        "verbs": [
          "get"
        ],
        "nonResourceURLs": [
          "/api1",
          "/api1/*",
          "/api2",
          "/api2/*"
        ]
      }
    ]
  }
```
How should we define the `visibilityRules` (or however it will be called) field in the configuration? Should it have the same structure as the object above?

PR #523, opened 2019-01-29
Determine if the user can see the admin views

dariadomagala:
**Description**

- Add a check to determine if the user can see the admin views

**Related issue(s)**

Review together with https://github.com/kyma-project/kyma/pull/2486


core/src/luigi-config/luigi-config.js
```diff
@@ -419,6 +421,87 @@ function fetchFromGraphQL(query, variables) {
...
+    );
+  });
+}
+
+function navigationPermissionChecker(nodeToCheckPermissionsFor) {
+  let hasPermissions = false;
+  if (nodeToCheckPermissionsFor.adminOnly) {
+    if (selfSubjectRulesReview.length > 0) {
+      selfSubjectRulesReview.forEach(rule => {
+        if (
```
pekura: OK, this condition is.. not the final solution I suppose. But can't we do it better? Like having a list of api groups and verbs that we consider admin must have (as a constant) and a generic function that checks again any list of verbs and api groups (that would know that '*' means any etc..). We can talk to Philipp first to make sure that this approach with selfsubjectrules is, in general, not temporary.

core/src/luigi-config/luigi-config.js
```diff
@@ -454,14 +537,24 @@ function relogin() {
...
+        modules.backendModules.forEach(backendModule => {
           backendModules.push(backendModule.name);
         });
       }
+      if (subjectRules && subjectRules.length > 0) {
+        selfSubjectRulesReview = subjectRules;
+      }
     },
     err => {
       console.error('Error while fetching backend modules', err);
```
pekura: Even if error cause is selfsubjectrules call the error message says:

luigiConfig.bundle.js:1 Error while fetching backend modules Forbidden (user=admin@kyma.cx, verb=create, resource=selfsubjectrulesreviews, subresource=).

core/src/luigi-config/luigi-config.js
dariadomagala: No, it's not the final solution as described here: https://github.com/kyma-project/console/issues/459#issuecomment-458522570 . This check was actually a proposal from Philipp for the temporary solution.

pekura: Alright then!

PR #537, opened 2019-02-04
Browser cache invalidation improvements

pekura:


parostatkiem:
Did you forget the PR?

pekura:
This is a quick solution for the 0.7 release. The combination of etags and cache max-age makes it a lot better and performance is acceptable. Maximum performance would be achieved by a version number or content hash based id as part of the name of each resource but this requires more time to discuss and implement - we might improve for the 0.8 release.

lambda/nginx/nginx.conf
```diff
@@ -39,6 +39,8 @@ http {
             root   /usr/share/nginx/html;
             index  index.html index.htm;
             try_files $uri$args $uri$args/ /index.html;
+
```
mjasinski5: In other views there's also 
`add_header Access-Control-Allow-Origin *;` to enable cors. Is this intended to omit this line?

brokers/nginx.conf
```diff
@@ -26,11 +26,13 @@ http {
...
     root /var/public;
     port_in_redirect off;
-    add_header Access-Control-Allow-Origin *;
 
     location / {
 
         try_files $uri$args $uri$args/ $uri $uri/ /;
+
+        add_header 'Cache-Control' 'public, max-age=300';
+        add_header Access-Control-Allow-Origin *;
```
mjasinski5: Is that intended to disable cors on /status endpoint?

lambda/nginx/nginx.conf
pekura: This header is not necessary for (microfrontend) views, I did not add or remove it here and tried not to touch it in the other views as well.

brokers/nginx.conf
pekura: No, that was not my intention, I did not want to change the behaviour for this header. I re-added it for /status endpoint, thanks for pointing it out!

PR #568, opened 2019-02-13
generic alerts for console

JohannesDoberer:


core/src/app/content/environments/environments-container/environments-container.component.ts
```diff
@@ -203,6 +204,25 @@ export class EnvironmentsContainerComponent implements OnInit, OnDestroy {
             limitExceededErrors.length > 0
           ) {
             this.setLimitExceededErrorsMessages(limitExceededErrors);
+            const linkdata = {
+              goToResourcesConfig: {
+                text: 'Resources Configuration',
+                url: `/home/namespaces/${env}/resources`
+              }
+            };
+            let errorText = `Error! The following resource quota limit has been exceeded by the given resource:<br><br>`;
```
dariadomagala: I'm not sure if double `<br>` tag is needed, one would be enough. This error is too big in my opinion.

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -280,7 +280,12 @@ export class LambdaDetailsComponent
     this.lambdaDetailsService
       .getResourceQuotaStatus(this.namespace, this.token)
       .subscribe(res => {
```
dariadomagala: If there is an error, the lambda is not created / updated, and the user doesn't get information why. I don't think we should let the user know about this error, but we should let him create the function anyway. So please move everything what is after a posmessage to `finally` handler.

rakesh-garimella:
The behavior not consistent. If I acknowledge the message by clicking `x`  and migrate to lambdas then I dont see this message any more(Looks like no graphql request is made). Also happens if I migrate to `api`, `resources` and `permissions` microfrontends.

JohannesDoberer:
> The behavior not consistent. If I acknowledge the message by clicking `x` and migrate to lambdas then I dont see this message any more(Looks like no graphql request is made). Also happens if I migrate to `api`, `resources` and `permissions` microfrontends.

I checked the behavior that you described and it is the same as it was before. We had the same behavior with 'old' alerts.

PR #584, opened 2019-02-15
csp for console

JohannesDoberer:


JohannesDoberer:
There are also changes for kyma https://github.com/kyma-project/kyma/pull/2870

PR #586, opened 2019-02-18
Add nosniff setting for MIME types on all console components

jesusreal:
**Description**

- Add [nosniff setting for MIME types](https://www.keycdn.com/support/what-is-mime-sniffing) on all console components

**Related issue(s)**


jesusreal:
Hi @pkosiec thanks for your comment. IMHO this is not a problem, we should always provide proper MIME type and avoid any security vulnerability that could be produced by allowing browser MIME sniffing feature.

PR #613, opened 2019-02-28
Role binding for users

JohannesDoberer:


core/src/app/shared/services/rbac.service.ts
```diff
@@ -18,28 +18,27 @@ export class RbacService {
...
+      name: `${data.name}-${data.roleName}-binding`
     };
   }
 
   prepareBindingToCreate(data) {
+    const subject = {
+      apiGroup: 'rbac.authorization.k8s.io',
+      name: data.name
+    };
+    subject['kind'] = data.isUserGroup ? 'Group' : 'User';
```
parostatkiem: Is there any reason why it is not defined inside `subject` object?

core/src/app/shared/components/role-binding-modal/role-binding-modal.component.ts
```diff
@@ -92,38 +97,39 @@ export class RoleBindingModalComponent implements OnDestroy {
...
-    }
-
-    return {
-      kind: 'RoleBinding',
-      groupName: this.userGroup,
+    const data = {
       roleKind: this.selectedKind,
       roleName: this.selectedRole,
-      namespace: this.currentEnvironmentId
+      name: this.userOrGroup
```
parostatkiem: You can just write ` isUserGroup: this.isUserGroup` instead of the definition below.
Also, is there any particular reason why you use `data['property']` instead of `data.property`?

core/src/app/shared/components/role-binding-modal/role-binding-modal.component.html
```diff
@@ -13,38 +15,49 @@ <h4 class="tn-modal__title">Create Binding</h4>
             <div class="tn-form__group">
               <div class="error-message-placeholder" *ngIf="error">{{ error }}</div>
 
+              <div class="fd-form__set">
+                <div class="fd-form__group">
+                  <div class="fd-form__item fd-form__item--inline fd-form__item--check">
+                    <label class="fd-form__label" for="pDidh767">
+                      <input type="radio" class="fd-form__control" id="pDidh767" name="radio"
+                        (click)='toogleUserGroup()' [checked]="!isUserGroup">
```
parostatkiem: Wouldn't it be better to set `isUserGroup` depending on currently selected radio button?
For now you can click any radio button to toggle this value which is unintuitive in my opinion.

core/src/app/shared/services/rbac.service.ts
JohannesDoberer: done

core/src/app/shared/components/role-binding-modal/role-binding-modal.component.ts
JohannesDoberer: >Also, is there any particular reason why you use `data['property']` instead of `data.property`?
If property is not defined in object, vscode will underline it (because property is not there).
I know dot notation also will work, but for me it is nicer.

parostatkiem: A discussable thing but OK 👍

Issue #590, opened 2019-02-19
Ci setup for OData React component


magicmatatjahu:
**Description**
After merged #588 PR:
- create Prow Job for OData React component with `presubmit` and `postsubmit` stage in `test-infra` repository.
- create Makefile with pipeline for pr and master like in other package in `console` repo.

Stretch:
- create bash script for automatically pushing new version of component to npm, try to find a solution for pushing patch, minor and major version, eg with appropriate tag in title of PR.

**Reasons**
Better development of OData React component.

aerfio:
[x] - also make `pre-master-kyma-components-asset-upload-service` required

PR #618, opened 2019-03-01
Add Makefile for react-odata component

aerfio:
**Description**

- Add `Makefile` for CI
- Fix leftover errors and bugs from last PR that prevent Makefile from working

**Related issue(s)**
See also #590, https://github.com/kyma-project/test-infra/pull/670


components/react-odata/Makefile
```diff
@@ -0,0 +1,20 @@
...
+	npm run build
+
+validate:
+	npm run conflict-check
+	npm run lint-check
+
+test:
+	CI=true npm run test
+
+scan:
```
michal-hudy: We don't use whitesource anymore ;)

PR #625, opened 2019-03-04
Add prettier in validation in Instances, Catalog, components/react, components/react-odata, brokers, content and tests

### LINKED ISSUE #628: Add prettier in validation in Instances, Catalog and components/react
**Description**

We should add prettier validation in order to have all of our codebase comply with our prettier rules

**Reasons**

Without prettier and other linters we edit files without clear reason, just by editing - with prettier check we also have less code to check in PRs

aerfio:
**Description**

Add prettier in validation in Instances, Catalog and components/react

- Add prettier validation in Instances, Catalog and components/react

**Related issue(s)**

Resolves #628


catalog/Makefile
```diff
@@ -15,7 +15,8 @@ build:
 
 validate:
 	eslint -c .eslinterrc ./src
-
+	npx prettier --config ../.prettierrc --check "./src/**/*.{js,jsx}"
```
pkosiec: Please validate `css`, `json`, `ts` and `tsx` in all projects

pkosiec: Maybe also `html`, `graphql` (if we have any) and `yaml`?

aerfio: We don't have any `.graphql` files, but I'll add rest of those extensions

aerfio:
All of the changed files in `catalog`, `instances` and `components/react` were changed by `prettier --write` btw

components/react/Makefile
```diff
@@ -9,6 +9,7 @@ build:
 	npm run build
 
 validate:
+	npm run lint-check
 	eslint -c .eslinterrc ./src
```
pkosiec: Can we run here TSLint validation, as all components are now in TypeScript?
Also it would be great to add it to `lint-check` script.

components/react/package.json
```diff
@@ -18,7 +18,9 @@
     "test": "jest src/**; exit 0",
     "test:watch": "jest src/** --watch",
     "test:watchAll": "jest src/** --watchAll",
-    "test:coverage:report": "open ./coverage/lcov-report/index.html"
+    "test:coverage:report": "open ./coverage/lcov-report/index.html",
+    "lint-fix": "prettier --config .prettierrc --write \"./src/**/*.{js,jsx,ts,tsx,html,css,json}\"",
+    "lint-check": "prettier --config .prettierrc --check \"./src/**/*.{js,jsx,ts,tsx,html,css,json}\""
```
pkosiec: Please add TSLint here in lint-check

components/react/Makefile
```diff
@@ -9,6 +9,7 @@ build:
 	npm run build
 
 validate:
+	npm run lint-check
```
pkosiec: Can we run here TSLint validation instead of eslint, as all components are now in TypeScript? It could be done as part of `npm run lint-check`

components/react/package.json
aerfio: There are no `.ts`/`.tsx` files in components/react

aerfio:
There are no `.ts`/.`tsx` files in components/react, only typescript component we have is react-odata, which already is checked with tslint

PR #676, opened 2019-03-11
Fix Safari bug with Popovers

parostatkiem:
**Description**
It looks like Safari tries to do some fictional `translateZ` transition on Popover's body.
I disabled it and it works fine. There are no visual changes (transition works like id did before) because there was actually no Z-axis change (Safari thinks different).

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/core/src/styles/fiori-fundamentals-overrides.scss
+++ b/core/src/styles/fiori-fundamentals-overrides.scss
@@ -171,5 +171,5 @@ a {
 // Fix for popover could not be opened on Safari (also causing some infinite operations eating all resources)
 // It should not provide any visual changes
 .fd-popover__body {
-  transition: all 0.135s, transformZ 0;
+  transition: all 0.135s, transformZ 0s;
 }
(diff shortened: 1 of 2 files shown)


pekura:
Wtf. Respect for finding it :-)

marynaKhromova:
@parostatkiem Hey :) I tried to find a solution to fix the issue and to keep the animation at the same time. The idea is to remove `transform` at all and to use `margin-top` instead. Here is a solution:

```
// Fix for popover could not be opened on Safari (also causing some infinite operations eating all resources)
// Transform property causes problem in Safari. Workaround removes `transform` and uses `margin-top` instead
.fd-popover__body  {
  margin-top: 0;
  transform: none;
}

.fd-popover__body {
  &[aria-hidden='true'], 
  &.is-hidden {
    margin-top: -15px;
    transform: none;
  }
}
```
Let me know if it makes any sence for you :)

JohannesDoberer:
👏

PR #680, opened 2019-03-12
Fix lambda modal close

maxmarkus:
Added check for modal instance existence since it might have been removed already by fiori ngx internal modal triggers.
Also includes missing css styles that got removed in the latest fiori release.

Fixes this issue:
Bildschirmfoto 2019-03-12 um 10 50 44

Also issues to be fixed:
- add `fd-has-display-block` helper class (it’s missed in last Fiori Fundamentals version)
it will fix the issues with small width dropdown

-  remove `./node_modules/fiori-fundamentals/dist/fiori-fundamentals.css` from angular.json because we use scss includes 
-  update Fiori Fundamentals to version 1.4.3


lambda/src/app/lambdas/lambda-details/lambda-instance-bindings/lambda-instance-binding-creator/lambda-instance-binding-creator.component.ts
```diff
@@ -110,9 +119,14 @@ export class LambdaInstanceBindingCreatorComponent {
   }
 
   public closeModal(event: Event): void {
+    if (!this.isActive) {
+      return;
+    }
     event.stopPropagation();
     luigiClient.uxManager().removeBackdrop();
-    this.modalService.close(this.instanceBindingCreatorModal);
+    if (this.instanceBindingCreatorModal) {
```
dariadomagala: I think this check is not needed

lambda/src/app/fetch-token-modal/fetch-token-modal.component.ts
```diff
@@ -40,6 +49,10 @@ export class FetchTokenModalComponent {
   }
 
   public cancel(event: Event) {
+    if (!this.isActive) {
```
dariadomagala: It would be better to have 
```
if (this.isActive) { 
  event.preventDefault();
  this.modalService.close(this.fetchTokenModal);
}
```
Same for each function in this PR that has this check

lambda/src/app/lambdas/lambda-details/event-trigger-chooser/event-trigger-chooser.component.ts
```diff
@@ -105,8 +104,10 @@ export class EventTriggerChooserComponent {
   }
 
   closeEventTriggerChooserModal() {
+    if (!this.isActive) {
+      return;
+    }
     this.isActive = false;
```
dariadomagala: I think we can remove this line from here, right?

lambda/src/app/fetch-token-modal/fetch-token-modal.component.ts
maxmarkus: Matter of taste, but fine for me.

PR #681, opened 2019-03-12
Hide 'expose api' action for system namespaces

dariadomagala:
**Description**

- Hide 'expose api' action for system namespaces

**Related issue(s)**


core/src/app/content/environments/operation/services/services-entry-renderer/services-entry-renderer.component.ts
```diff
@@ -17,22 +17,37 @@ export class ServicesEntryRendererComponent
     private componentCommunicationService: ComponentCommunicationService
   ) {
     super(injector);
+    this.contextListenerId = LuigiClient.addContextUpdateListener(context => {
```
maxmarkus: This might lead to a race condition. I would suggest to include _this.actions_ to the contextUpdateListener callback.

PR #689, opened 2019-03-14
Use graphql for pods

dariadomagala:
**Description**

- Use graphql for pods
- Refactor AbstractGraphqlElementListComponent so it can be used for resources other than replica sets

**Related issue(s)**

### DIFF
--- a/core/src/app/content/environments/operation/pods/pods-entry-renderer/pods-entry-renderer.component.spec.ts
+++ b/core/src/app/content/environments/operation/pods/pods-entry-renderer/pods-entry-renderer.component.spec.ts
@@ -21,13 +21,7 @@ describe('PodsEntryRendererComponent', () => {
             useValue: {
               name: 'name',
               status: 'RUNNING',
-              containerStates: [
-                {
-                  state: 'RUNNING',
-                  reason: '',
-                  message: ''
-                }
-              ]
+              containerStates: [{}]
             }
           }
         ],
@@ -48,23 +42,7 @@ describe('PodsEntryRendererComponent', () => {
     expect(component).toBeTruthy();
   });
 
-  it('should display pod status if 1 out of 1 container failed', () => {
-    const pod = {
-      status: 'PENDING',
-      containerStates: [
-        {
-          state: 'PENDING',
-          reason: 'ImagePullBackOff',
-          message: 'Image not found'
-        }
-      ]
-    };
-    expect(component.getStatus(pod)).toEqual(
-      'PENDING: ImagePullBackOff'
-    );
-  });
-
-  it('should display pod status if 1 out of 2 containers failed', () => {
+  it('should display pod status any container failed', () => {
     const pod = {
       status: 'PENDING',
       containerStates: [
@@ -86,16 +64,10 @@ describe('PodsEntryRendererComponent', () => {
 
   });
 
-  it('should display pod status if 1 out of 2 containers failed', () => {
+  it('should display pod status if it\'s running', () => {
     const pod = {
       status: 'RUNNING',
-      containerStates: [
-        {
-          state: 'RUNNING',
-          reason: '',
-          message: ''
-        }
-      ]
+      containerStates: [{}]
     };
     expect(component.getStatus(pod)).toEqual(
       'RUNNING'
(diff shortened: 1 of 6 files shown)


core/src/app/content/environments/operation/pods/pods.component.ts
```diff
@@ -1,71 +1,56 @@
...
   }
 
-  public ngOnDestroy() {
-    this.currentEnvironmentSubscription.unsubscribe();
+  getGraphglQueryForList() {
+    return `query Pod($namespace: String!) {
+      pods(namespace: $namespace) {
+        name
+        nodeName
+        namespace
```
jesusreal: Maybe we don't need to retrieve the `namespace` and `json` fields? I don't see them being used and that would be more efficient.

core/src/app/content/environments/operation/pods/pods.component.ts
```diff
@@ -1,71 +1,56 @@
...
     changeDetector: ChangeDetectorRef
   ) {
-    super(currentEnvironmentService, changeDetector, http, commService);
-    const converter: DataConverter<IPod, Pod> = {
-      convert(entry: IPod) {
-        return new Pod(entry);
-      }
-    };
-
-    this.currentEnvironmentSubscription = this.currentEnvironmentService
```
jesusreal: Looks like `AbstractGraphqlElementListComponent` is gonna help us remove a lot of duplicated code for GraphQL API usage. Great! :)

core/src/app/content/environments/operation/pods/pods-entry-renderer/pods-entry-renderer.component.ts
```diff
@@ -114,8 +76,8 @@ export class PodsEntryRendererComponent
   }
 
   hasErrors(entry) {
-    return entry.status.containerStatuses.some(
-      status => status.state[Object.keys(status.state)[0]] === 'running'
+    return entry.containerStates.some(
+      status => status.state !== 'RUNNING' && status.message
```
jesusreal: Even if not added by you, let´s don't call it `status`, since this gets confused with entry.status. As you are applying a method on an Array, for me is enough is you just use `s` instead of `status`.

core/src/app/content/environments/operation/pods/pods-entry-renderer/pods-entry-renderer.component.spec.ts
```diff
@@ -49,89 +48,65 @@ describe('PodsEntryRendererComponent', () => {
...
-      'Stopped (Signal: Some signal)'
+    expect(component.getStatus(pod)).toEqual(
+      'PENDING: ImagePullBackOff'
     );
+  });
 
-    const exitCodeStatus = {
-      phase: 'Doing something..',
-      containerStatuses: [
+  it('should display pod status if 1 out of 2 containers failed', () => {
```
jesusreal: IMHO this test makes the test right above irrelevant, and we could remove that one. What do you think?

core/src/app/content/environments/operation/pods/pods-entry-renderer/pods-entry-renderer.component.spec.ts
```diff
@@ -49,89 +48,65 @@ describe('PodsEntryRendererComponent', () => {
...
 
-    const otherCodeStatus = {
-      phase: 'Doing something..',
-      containerStatuses: [
+  });
+
+  it('should display pod status if 1 out of 2 containers failed', () => {
+    const pod = {
+      status: 'RUNNING',
+      containerStates: [
```
jesusreal: We can simplify "containerStates" to `containerStates: [ {}, {} ]`, makes the test smaller and easier to understand

core/src/app/content/environments/operation/pods/pods-entry-renderer/pods-entry-renderer.component.spec.ts
```diff
@@ -49,89 +48,65 @@ describe('PodsEntryRendererComponent', () => {
...
-    };
-    expect(component.getStatus({ status: noContainerStatesStatus })).toEqual(
-      'Doing something..'
+    expect(component.getStatus(pod)).toEqual(
+      'RUNNING'
     );
   });
 
   it("should disable the pod if 'disable' event with right data has been sent", async done => {
     fixture.detectChanges();
```
jesusreal: this detectChanges call is not needed

PR #694, opened 2019-03-14
Proper usage of anchor tags in console UI

marynaKhromova:
Provided changes:
- set `href="javascript:void(null)"`  for links where no redirection takes place
- remove unnecessary styles
- tiny fix for table


lambda/src/styles/fiori-fundamentals-overrides.scss
```diff
@@ -28,6 +28,21 @@
...
+  .disabled {
+    opacity: 0.5;
+    pointer-events: none;
+  }
+
+  td {
+    vertical-align: top;
+  }
+  
+  tr {
```
dariadomagala: Is it needed?

PR #700, opened 2019-03-19
Log ui for lambdas and pods

pekura:


pekura:
To test it locally:
- checkout kyma from this branch: https://github.com/kyma-project/kyma/pull/3240/files
- apply this patch [patch.diff.zip](https://github.com/kyma-project/console/files/2982742/patch.diff.zip) to kyma and run it

PR #702, opened 2019-03-20
Clear binding modal data after closing it

dariadomagala:
**Description**

- Clear binding modal data after closing it

**Related issue(s)**


core/src/app/shared/components/role-binding-modal/role-binding-modal.component.ts
```diff
@@ -107,16 +107,22 @@ export class RoleBindingModalComponent implements OnDestroy {
     }
     this.modalService.open(this.createBindingModal).result.finally(() => {
       this.isActive = false;
+      this.clearData();
     });
     this.isUserGroupMode = true;
   }
 
   public close() {
+    this.clearData();
```
jesusreal: This is not needed, since the `finally` block is always executed when we close the modal ;)

PR #720, opened 2019-03-26
Improve login error behavior

maxmarkus:
Added loginError state handling to logout.html 
To be tested in combination with Luigi PR


core/src/logout.html
```diff
@@ -62,8 +62,30 @@
...
+switch (reason) {
+  case 'tokenExpired':
+    document.getElementById('headline').innerText = 'Your session has expired.';
+    break;
+  case 'loginError':
+    document.getElementById('headline').innerText = 'There was an error while logging in.';
+    document.getElementById('message').innerText = 'Please try to log in again.';
+    break;
+  default:
+    break;
```
parostatkiem: How about some generic error like `Ooops, something went wrong. Try to log in again`?

core/src/logout.html
```diff
@@ -62,8 +62,30 @@
...
+  var regex = new RegExp('[\\?&]' + name + '=([^&#]*)');
+  var results = regex.exec(location.search);
+  return results === null ? '' : decodeURIComponent(results[1].replace(/\+/g, ' '));
+};
+
+var reason = getUrlParameter('reason');
+var error = getUrlParameter('error');
+
+if (error) {
+  document.getElementById('message').innerText = error;
```
parostatkiem: In case we have error with reason 'loginError', the error won't be visible. Am I wrong?

core/src/logout.html
```diff
@@ -62,8 +62,30 @@
   window.location.href = window.location.origin;
 }
 
-if (window.location.href.includes('tokenExpired')) {
-  document.getElementById('logout-headline').innerText = 'Your session has expired.';
+function getUrlParameter(name) {
+  name = name.replace(/[\[]/, '\\[').replace(/[\]]/, '\\]');
+  var regex = new RegExp('[\\?&]' + name + '=([^&#]*)');
```
parostatkiem: why `var`? It's not used outside...

core/src/logout.html
```diff
@@ -62,8 +62,30 @@
...
+  var results = regex.exec(location.search);
+  return results === null ? '' : decodeURIComponent(results[1].replace(/\+/g, ' '));
```
parostatkiem: Why not `results ? ...`?

core/src/logout.html
bszwarc: This one sounds fine to me :)

core/src/logout.html
maxmarkus: legacy browser compatibility. it is available from ie11 on, which is the last one we support. so we could refactor to use let / const too.

core/src/logout.html
maxmarkus: Yes, this is expected.

core/src/logout.html
parostatkiem: I think I don't get it. You said IE11 supports it (which I confirm by researching it myself). So why do you still want to use `var`?

PR #728, opened 2019-03-28
673 make all console views more consistent

marynaKhromova:

- create constant for empty text
- make empty fields consistent 
- make same order of `+Add ..` and `Search` buttons
- implement `Search` and `Filter` buttons as icons (without text)
- change design of current filter counter (the design has been discussed with our designers)


core/src/app/content/namespaces/operation/services/service-details/service-details.component.ts
```diff
@@ -21,6 +22,7 @@ export class ServiceDetailsComponent implements OnInit, OnDestroy {
   public serviceDetailsUrl: string;
   public currentNamespaceSubscription: Subscription;
   public isSystemNamespace: boolean;
+  private emptyText = EMPTY_TEXT;
```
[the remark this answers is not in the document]
parostatkiem: I confirm there will be errors on CI (and by `npm run build` also).

core/src/app/shared/constants/constants.ts
```diff
@@ -0,0 +1 @@
+export const EMPTY_TEXT = '-';
```
parostatkiem: Wouldn't it be better to have this on root level? As @jesusreal mentioned, this also needs to be used in Lambdas. Furthermore, in the near future we are going to take over all Console components which means this file would have to be duplicated `n` times.

core/src/app/content/namespaces/configuration/apis/apis.component.html
```diff
@@ -4,15 +4,15 @@
...
-                Create API
-            </button>
             <y-list-search [hidden]="!initialized || showEmptyPage" [filterState]="filterState"
                 (onFilterChanged)="onFilterChanged($event)">
             </y-list-search>
             <y-list-filter [hidden]="!initialized || showEmptyPage || hideFilter" [filterState]="filterState"
                 (onFilterChanged)="onFilterChanged($event)">
             </y-list-filter>
+            <button fd-button (click)="navigateToCreate()" [options]="'light'" [glyph]="'add'">
+                Create API
```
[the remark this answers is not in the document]
marynaKhromova: Here the changes were about the order of the buttons. `+` button should be everywhere with a text according to the design.

core/src/app/shared/constants/constants.ts
jesusreal: I hope I understand correctly what you mean. Lambda is a different application than core and if you add it to the root and add the file somehow to the build of each application, this is a design topic which need some discussion IMHO.

parostatkiem: hmm, you are right. I just thought about the need of duplicating it many times. However, only Core and Lambda are going to use it (which I forgot previously) so the file will be duplicated only once which is not a big deal 👌

Issue #742, opened 2019-04-03
Use "simpleMobileOnly" responsive navigation setting


hardl:
newest luigi version has an option to slide in left navigation automatically on mobile devices (width<600px)

Activate in console luigi config

```
...
settings: {
    responsiveNavigation: 'simpleMobileOnly',
    ...
}

```

marynaKhromova:
Will be implemented within this task https://github.com/kyma-project/console/issues/762

PR #743, opened 2019-04-03
#742 Use "simpleMobileOnly" responsive navigation setting

hardl:
Use "simpleMobileOnly" responsive navigation setting.

changes are only effective after update to newest luigi rc

### DIFF
--- a/core/src/index.html
+++ b/core/src/index.html
@@ -3,6 +3,7 @@
 
 <head>
   <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1, maximum-scale=1, minimum-scale=1">
   <title>Kyma</title>
   <base href="/">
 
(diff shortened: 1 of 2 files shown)


core/src/index.html
```diff
@@ -3,6 +3,7 @@
 
 <head>
   <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1, maximum-scale=1, minimum-scale=1">
```
marynaKhromova: With this tag, we kind of provide the responsiveness, and since now the view is not just scaled, you can see in many places not perfectly styled layouts (title is cut off, block width is too small and etc.). Shouldn't we fix those issues before we release it?

Some examples here:

core/src/luigi-config/luigi-config.js
```diff
@@ -831,6 +831,7 @@ Promise.all([getBackendModules(), getSelfSubjectRulesReview(), getFreshKeys()])
         skipRoutingForUrlPatterns: [/access_token=/, /id_token=/]
       },
       settings: {
+        responsiveNavigation: 'simpleMobileOnly',
```
marynaKhromova: Just a question. Do we have to do the same in our Angular example in luigi?

hardl:
We agreed that we first need to look in the above issues

Issue #713, opened 2019-03-25
Catalog and Instance views to support Headless CMS


derberg:
**Description**

- refactor content to use react-markdown to render markdown into html in real time
- implement logic to fetch markdown files provided in the graphql queries for DocsTopic resource
- backward compatible with old solution, Headless CMS is default one

**Reasons**

Support new way of rendering markdown and specs that are shared witn new Headless CMS

PR #745, opened 2019-04-04
Catalog and Instance UI to support Headless CMS

aerfio:
**Description**

- add fetching mechanism for docs from docsTopics in Catalog and Instances UI, which is default mechanism of presenting docs
- existing solution will be fallback, using deprecated apis

**Related issue(s)**

See #713


catalog/src/components/ServiceClassDetails/ServiceClassTabs/ServiceClassTabs.component.js
```diff
@@ -1,104 +1,330 @@
...
-                theme={asyncApiTheme}
-                config={asyncApiConfig}
-              />
+
+      const deprecatedDocs =
+        documentsTypes &&
+        documentsTypes.map(type =>
+          documentsByType &&
+          documentsByType[type] &&
+          validatDocumentsByType(documentsByType[type]) ? (
```
akucharska: if(!...){ 
return null;
}
return  <Tab ...>

components/react/src/components/ReactMarkdown/components/Theme.js
```diff
@@ -0,0 +1,103 @@
+export default {
+  plain: {
+    backgroundColor: 'rgb(250, 250, 250)',
+    color: 'rgb(11, 116, 222)',
+  },
+  styles: [
+    {
+      types: ['string'],
```
akucharska: This is older version of file

aerfio:
One thing I want to discuss is whether to leave `console.error` in 
```
   if (error) {
      console.error(error);
      return <div>{`${error.name}: ${error.message}`}</div>;
    }
```

aerfio:
both instances and catalog-ui

PR #747, opened 2019-04-04
Keep existing API in lambda configuration

### LINKED ISSUE #724: API is overwritten on Update
**Description**

If a Lambda API is created with a not default authentication provider, the IDP Settings are overwritten on every lambda update using lambda-ui

**Expected result**

API Object is not changed, if nobody touched it.

**Actual result**

Settings are overwritten with the default IDP settings.

**Steps to reproduce**

1. Create Lambda with HTTP Trigger and use a not default IDP Settings
2. Do changes to the lambda function
3. Check the API Object configuration. It is changed.

**Troubleshooting**

none

k15r:
**Description**

- existing API configuration was overwritten by default values
**Related issue(s)**

Fixes #724

### DIFF
--- a/lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
+++ b/lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
@@ -53,6 +53,7 @@ import { Subscription } from '../../shared/datamodel/k8s/subscription';
 import { SubscriptionsService } from '../../subscriptions/subscriptions.service';
 import { EventTriggerChooserComponent } from './event-trigger-chooser/event-trigger-chooser.component';
 import { HttpTriggerComponent } from './http-trigger/http-trigger.component';
+import { Authentication } from '../../shared/datamodel/authentication';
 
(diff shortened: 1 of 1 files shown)


lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -1174,6 +1177,9 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
     if (api.spec.authentication !== undefined) {
       httpEndPoint.isAuthEnabled =
         api.spec.authentication.length !== 0 ? true : false;
+        if (httpEndPoint.isAuthEnabled) {
+          httpEndPoint.authentication = api.spec.authentication[0]
```
sayanh: Incase you have more than 1 AuthenticationRule configuration, this won't work.

sayanh: As discussed with @joek, this is all good.

Issue #758, opened 2019-04-08
Fetch user permissions via qgl


kwiatekus:
We need a way to fetch an information  whether user is an admin user or not

PR #771, opened 2019-04-11
Evaluating self subject rules

kwiatekus:
**Description**

- Find matching rbac rules for navigation nodes

**Related issue(s)**
https://github.com/kyma-project/console/issues/758


core/src/luigi-config/luigi-config.js
```diff
@@ -775,7 +813,12 @@ Promise.all([getBackendModules(), getSelfSubjectRulesReview(), getFreshKeys()])
                         ]
                       }
                     ],
-                    adminOnly: true
+                    // adminOnly: true,
+                    requiredPermissions : [{
+                      apiGroup : "rbac.authorization.k8s.io",
+                      resource : "clusterrolebindings1",
+                      verb : "create"
```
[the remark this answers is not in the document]
kwiatekus: hmm.. I think I cant do it having 3 arrays.. but indeed an array for verbs is a good idea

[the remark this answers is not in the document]
kwiatekus: What if the node requires:
- "list", "create" and "delete" for one resource A from apiGroup A 
- "list" only from resource B from apiGroup B.

I couldnt model it as one `requiredPermission` entry, i.e : 
```
requiredPermissions : [{
                      apiGroups : ["apiGroupA","apiGroupB"],
                      resources : ["resourceA","resourceB"],
                      verbs : ["create", "delete", "list"]
}]
```
I would really need to separate it on resource level - so two items in the array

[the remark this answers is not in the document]
dariadomagala: Yes, you would need two arrays, like 
```
[{
  apiGroups : ["apiGroupA"],
  resources : ["resourceA"],
  verbs : ["create", "delete", "list"]
}, {
  apiGroups : ["apiGroupB"],
  resources : ["resourceB"],
  verbs : ["list"]
}]
```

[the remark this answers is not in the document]
dariadomagala: Not sure if `apiGroups` needs to be an array, you might be right to just have it as a string - probably will be easier to implement it.

core/src/luigi-config/luigi-config.js
```diff
@@ -560,25 +560,63 @@ function getSelfSubjectRulesReview() {
...
+    });
   }
   return hasPermissions;
 }
 
+function arrayContainStringOrJoker(array, stringToFind, jokerString) {
+    if(array === null || array === undefined) {
+      return false;
+    }
+    let contains = false
```
jesusreal:
```
    return Boolean(array.find(x => x === stringToFind || x === jokerString))
```
would do the same in one line ;)

jesusreal: Same applies below for `matchFound`

core/src/luigi-config/luigi-config.js
```diff
@@ -560,25 +560,63 @@ function getSelfSubjectRulesReview() {
...
+    });
+    return contains;
+}
+
+function matchingRuleFound(allrules, requiredRule) {
+  if(allrules === null || allrules === undefined) {
+    return false;
+  }
+  let matchFound = false
+  allrules.forEach(rule => {
```
jesusreal: Here maybe you want to use three times `filter` instead of `find` ;)

core/src/luigi-config/rbac-rules-matcher.js
```diff
@@ -0,0 +1,74 @@
...
+    !isNonEmptyArray(requiredPermissions) ||
+    !isNonEmptyArray(selfSubjectRulesReview)
+  ) {
+    return true;
+  }
+  for (var i = 0, len = requiredPermissions.length; i < len; i++) {
+    let requiredPermission = requiredPermissions[i];
+    for (var j = 0, vlen = requiredPermission.verbs.length; j < vlen; j++) {
+      let atomicVerb = requiredPermission.verbs[j];
+      let atomicVerbPermission = {
```
maxmarkus: Those three should be `const` since they get not changed later. And use let inside the `for` loops.

maxmarkus: Just have seen there are thousands of `var` in that file ;-)

Issue #773, opened 2019-04-12
'Create' buttons are clickable when they should be disabled


dariadomagala:
**Description**

'Create' buttons in IDPPresets and Application creations are clickable even if the form is empty. They look disabled but you can click on them.

**Expected result**

User shouldn't be able to click the 'create' button if the data he entered is wrong or incomplete. 

**Actual result**

User can click the 'create' button in IDPPresets or Applications modals with empty fields.
![Screen Shot 2019-04-12 at 08.34.21.png](https://images.zenhubusercontent.com/5b8399bea1576a587faa4a80/1fd19b43-f556-42e7-9739-ebd3f675d5e5)

**Steps to reproduce**

* Click on 'Create Application' in Applications view.
* Don't put any data to the inputs.
* Click 'Create' button.

**Troubleshooting**

PR #779, opened 2019-04-15
create button: avoid save when button state is disabled

maxmarkus:
save is not triggered when clicking _create_ on incomplete form.


maxmarkus:
Issue connection got lost, it was ticket #773

PR #784, opened 2019-04-23
Use websockets for pods and applications in console core

dariadomagala:
**Description**

- Use subscriptions in console core

**Related issue(s)**


core/src/app/content/namespaces/operation/deployments/deployments.component.spec.ts
```diff
@@ -2,16 +2,26 @@ import { ListModule } from 'app/generic-list';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { DeploymentsComponent } from './deployments.component';
 import { AppModule } from '../../../../app.module';
+import * as LuigiClient from '@kyma-project/luigi-client';
 import { APP_BASE_HREF } from '@angular/common';
 
+const mockLuigiClient = {
```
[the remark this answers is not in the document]
dariadomagala: It was failing at the beginning with few errors, one of them was about Luigi. Probably fixing other issue fix this one as well :) but good catch! 👍

core/src/app/content/namespaces/operation/graphql-data-provider.ts
```diff
@@ -30,16 +32,35 @@ export class GraphQLDataProvider implements DataProvider {
...
-      if (noCache || !this.queryCache) {
-        this.queryCache = this.graphQLClientService
-          .request(this.url, this.query, this.variables)
-          .pipe(
-            publishReplay(1),
-            refCount()
-          );
+
+      if(!this.subscriptions || ! this.resourceKind) {
+        if(noCache || !this.resourceQuery) {
```
jesusreal: Do we need to use the "noCache" flag in the else block corresponding to `if(!this.subscriptions || ! this.resourceKind) {`

core/src/app/content/namespaces/operation/graphql-data-provider.ts
```diff
@@ -6,20 +6,22 @@ import {
...
-import { delay, publishReplay, refCount } from 'rxjs/operators';
+import gql from 'graphql-tag';
+import { catchError, map } from 'rxjs/operators';
+import { GraphQLClientService } from 'shared/services/graphql-client-service';
 
 export class GraphQLDataProvider implements DataProvider {
   filterMatcher = new SimpleFilterMatcher();
   facetMatcher = new SimpleFacetMatcher();
-  queryCache: Observable<any>;
+  resourceQuery: any;
```
jesusreal: Let´s improve typing here by adding: QueryRef<any>; instead of any

core/src/app/content/namespaces/operation/graphql-data-provider.ts
```diff
@@ -30,16 +32,35 @@ export class GraphQLDataProvider implements DataProvider {
...
+          updateQuery: (prev, {subscriptionData}) => {
+            this.updateSubscriptions(prev, subscriptionData);
+          }
+        });
       }
 
-      this.queryCache.subscribe(
+      this.resourceQuery.valueChanges
+      .pipe(
+        map(res => {
```
jesusreal: Let´s do these ones one-line functions ;)

core/src/app/content/namespaces/operation/graphql-data-provider.ts
```diff
@@ -30,16 +32,35 @@ export class GraphQLDataProvider implements DataProvider {
...
+          document: gql`${this.subscriptions}`,
+          variables: this.variables,
+          updateQuery: (prev, {subscriptionData}) => {
+            this.updateSubscriptions(prev, subscriptionData);
+          }
+        });
       }
 
-      this.queryCache.subscribe(
+      this.resourceQuery.valueChanges
```
jesusreal: Maybe would make sense to save the subcription in a class variable and when creating a new subscription deleting the old one. Currently we create new subscriptions all the time when searching, filtering etc and this could be a performance issue.

kwiatekus:
Doesnt work on GCP cluster

Issue #793, opened 2019-04-25
Console sometimes makes additional http requests


dariadomagala:
**Description**

Sometimes console makes additional calls that are not supposed to be made in this specific scenario.

**Expected result**

Only necessary calls should be made and they should always behave in the same way.

**Actual result**

Sometimes after for example deleting a Pod, console sends requests to delete the Pod, and to get the new list of Pods (which are expected) but also sends the request to get the list of Services and Applications (not expected).

**Steps to reproduce**

1. In this scenario everithing works perfectly fine, only two calls are made:
* Open console in the Pods view
* Refresh the page
* Delete a Pod

2. Additional calls are made after deleting a pod in this scenario:
* Open console in the Namespace overview
* Navigate to the Pods view
* Delete a Pod

**Troubleshooting**

The issue after deleting a pod might be caused by subscriptions from the overview component that are still active (calls for Applications and Services are made in the overview page).

PR #799, opened 2019-04-29
Fix unnecessary requests for services and applications

y-kkamil:
**Description**

- Fixes unnecessary requests for services and application (caused by active subscriptions from abstract list component)
- Fixes unnecessary refresh request for pods and application, as it is covered by websockets implementation in #784 
- Removes memory leaks from abstract graphql list and abstract kubernetes list components
- Moves `subscribeToRefreshComponent` functionality out of abstract kubernetes list component to particular lists, as GraphQL-based components will not need it

**Related issue(s)**
See #793


y-kkamil:
#784 should be merged first, to avoid lack of refresh on pods & applications lists.

Issue #610, opened 2019-02-28
Split luigi-config.js and write unit tests


jesusreal:
Now that we have transpilation in place for our luigi configuration in console, it is time to take advantage of it.

_luigi-config.js_ is way too long currently and it would be good to have unit tests for (at least) some of the business logic contained in this file.

[x] Split luigi config file in several files
[x] Write unit test for important business logic

Please work on a timebox and focus in writing the most relevant tests, not everything needs to be tested.

kwiatekus:
We try to avoid such items in backlog. This doesn't bring value to the end-user.
We will do it as part of continuous refactoring of the code.

PR #829, opened 2019-05-10
Use view group caching in the console

pekura:


core/src/luigi-config/microfrontend-converter.js
```diff
@@ -109,7 +109,7 @@ function getDirectChildren(parentNodeSegments, spec, config) {
     });
 }
 
-function convertToNavigationTree(name, spec, config, segmentPrefix) {
+function convertToNavigationTree(name, spec, config, navigation, consoleViewGroupName, segmentPrefix) {
```
y-kkamil: this thing (and probably not only this) probably requires proper testing, see #610

PR #852, opened 2019-05-15
Fix bullet points in note/catution/tip

aerfio:
**Description**

- Fix bullet points in note/catution/tip

**Related issue(s)**

See also https://github.com/kyma-project/kyma/issues/3982


aerfio:
files changed in 2nd commit will be reverted later, they are only example for testing person

catalog/src/commons/helpers.js
```diff
@@ -125,6 +125,10 @@ export class DocsProcessor {
       if (doc.source.search(assetsRegexp) !== -1) {
         doc.source = doc.source.replace(assetsRegexp, `](${docsUrl}/assets`);
       }
+      doc.source += `
```
magicmatatjahu: Remove it

instances/src/commons/helpers.js
```diff
@@ -121,6 +121,11 @@ export class DocsProcessor {
       if (doc.source.search(assetsRegexp) !== -1) {
         doc.source = doc.source.replace(assetsRegexp, `](${docsUrl}/assets`);
       }
+
+      doc.source += `
```
magicmatatjahu: Remove it

content/src/components/DocsContent/DocsProcessor.js
```diff
@@ -25,6 +25,10 @@ export class DocsProcessor {
     this.docs.map(doc => {
       docsBeginingIndex = doc.source.indexOf(metadataRegexp, 3) + 3;
       doc.source = doc.source.substring(docsBeginingIndex, doc.source.length);
+      doc.source += `
```
magicmatatjahu: Remove it

components/react/src/components/ReactMarkdown/components/Blockquote.js
```diff
@@ -16,13 +16,30 @@ function BlockQuote({ children }) {
   };
 
   const createPanels = elem =>
```
magicmatatjahu:
``` js
const createPanels = elem => {
  if (!elem) return null;
  return elem.map((element, index) => (
    ...
  )
}
```

this is more readable.

components/react/src/components/ReactMarkdown/components/Blockquote.js
```diff
@@ -16,13 +16,30 @@ function BlockQuote({ children }) {
...
+      ? null
+      : elem.map((element, index) => (
+          <NotePanel type={getPanelType(element[0])} key={index}>
+            {element}
+          </NotePanel>
+        ));
 
-  return children ? <Fragment>{createPanels(children)}</Fragment> : null;
-}
+  const isOneOfTypes = arg => !!arg && ['note', 'caution', 'tip'].includes(arg);
```
magicmatatjahu:
``` js
const panelTypes = ['note', 'caution', 'tip'];
const isOneOfTypes = (arg: string) => Boolean(arg) && panelTypes.includes(arg);
```

components/react/src/components/ReactMarkdown/components/Blockquote.js
```diff
@@ -16,13 +16,30 @@ function BlockQuote({ children }) {
...
+      if (isOneOfTypes(currType)) {
+        return [...accumulator, [curr]];
+      }
+      const len = accumulator.length - 1;
+      const newLastElement = [...accumulator[len], curr];
+      return [...accumulator.slice(0, len), newLastElement];
+    }, []);
+
+  return modifiedChildren ? (
+    <Fragment>{createPanels(modifiedChildren)}</Fragment>
```
magicmatatjahu: Please do not use Fragment if you don't need pass props
```js
<>{createPanels(modifiedChildren)}</>
```

aerfio: Can't do without modifing babel config, better to leave it be until we'll do rewrite

aerfio:
Update of react was to meet missing peerDeps, removal of `prism-sth` and `react-markdown` was because those packages were in deps and devDeps simultaneously, so I've deleted them them from devDeps

PR #867, opened 2019-05-17
Lambda "Test" tab

parostatkiem:
**Description**

- add "Testing" tab to Lambda details
- display JSON editor in "test" tab
- add button to "call" Lambda over HTTPS using custom payload provided by user
- handle possible errors
- in case of success, inform user what he can do next
- add some info about _how to make my lambda testable_

**Related issue(s)**


lambda/src/app/lambdas/lambda-details/lambda-details.component.html
```diff
@@ -18,237 +18,295 @@
...
 
-  <app-fetch-token-modal #fetchTokenModal></app-fetch-token-modal>
-  <app-event-trigger-chooser #eventTriggerChooserModal (eventEmitter)="handleEventEmitter($event)">
-  </app-event-trigger-chooser>
-  <app-http-trigger #httpTriggerModal (httpEmitter)="handleHttpEmitter($event)"></app-http-trigger>
-</ng-container>
+      </fd-panel-body>
+      <fd-panel-footer>
+        <button fd-button [options]="'emphasized'" [glyph]="'journey-depart'" (click)="handleTestButtonClick()">
+          Make a test call
```
dariadomagala: Maybe just `test lambda`?

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -140,13 +140,19 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
...
   public jwksUri: string;
   public authType: string;
 
   public canShowLogs = false;
 
+  public currentTab = 'config';
+
+  public testPayloadText = JSON.stringify(this.testPayload, null, 2);
+
+
```
dariadomagala: Please remove those additional lines here as well :)

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -1293,4 +1305,42 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
...
+      body: this.testPayloadText,
+      headers: lambdaEndpoint.isAuthEnabled ? new Headers({
+        'Content-Type': 'application/json',
+        'Authorization': `Bearer ${this.token}`
+      }) : {}
+    }).finally(
+      () => { luigiClient.uxManager().hideLoadingIndicator(); }
+    ).then(res => {
+      if (!res.ok) { throw new Error(); }
+      luigiClient.uxManager().showAlert({ text: `The Lambda received your payload. You can now browse the logs to see the reaction`, type: 'info', closeAfter: 3500 });
```
dariadomagala: `the reaction` is not the best word fot that

lambda/src/app/lambdas/lambda-details/lambda-details.component.html
```diff
@@ -18,237 +18,295 @@
...
-                    ">Expose via HTTPS</a>
-                </li>
-              </ul>
-            </nav>
+
+        </fd-panel-filters>
+        <fd-panel-body>
+          <div *ngIf="!existingHTTPEndpoint" class="fd-alert fd-alert--information fd-has-margin-bottom-small"
+            role="alert">
+            You'll be able to <strong>test</strong> your lambda once it's deployed and has an HTTPS endpoint.
```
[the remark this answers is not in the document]
parostatkiem: I'm fine with the first proposal.
2nd and 3rd conflict with the idea of "disabling" the tab completely, which doesn't mean they are bad. 
I'll apply the first proposal but we can still discuss it in the team.

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -1293,4 +1305,42 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
...
+
+  handleTestButtonClick() {
+    if (!this.existingHTTPEndpoint) {
+      throw new Error('It looks like the Lambda is not deployed yet');
+    }
+
+    try {
+      this.testPayload = JSON.parse(this.testPayloadText);
+    } catch (ex) {
+      luigiClient.uxManager().showAlert({ text: `Couldn't parse payload JSON`, type: 'error', closeAfter: 3000 });
```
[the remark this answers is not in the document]
parostatkiem: good idea but think about the performance. The JSON.parse would have been called with every change of the payload (each keypress in fact).

lambda/src/app/lambdas/lambda-details/lambda-details.component.html
parostatkiem: I think it would be better to let the developer know *how* the lambda is actually tested - via HTTPS call. Mysterious `test lambda` could hide anything under the hood.

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
parostatkiem: I'm opened for suggestions about all these messages :) 
Any idea?

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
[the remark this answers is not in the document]
parostatkiem: I changed it to "local" alert inside Lambda MF, as you suggested.

sjanota:
I see one pretty annoying thing. Every time I make a change in lambda code I have to click "Save". It gets me back to lambda list then. However, I'm developing my lambda now, I have to add one more `console.log` to debug stuff. I don't want to have to navigate back to my lambda every time I make a change in code.

parostatkiem:
@sjanota seems pretty annoying but definitely not in scope of this task ;)

parostatkiem:
@kwiatekus fixed

lambda/src/app/lambdas/lambda-details/lambda-details.component.html
```diff
@@ -18,237 +18,310 @@
...
+
+    <fd-panel *ngIf='testingResult' class="y-fd-panel fd-has-margin-top-small">
+        <fd-panel-header>
+          <fd-panel-head>
+            <fd-panel-title>Response</fd-panel-title>
+          </fd-panel-head>
+        </fd-panel-header>
+  
+        <fd-panel-body>      
+          <ace-editor [readOnly]="true" id="test-result" [(text)]="testingResult" [theme]="theme"
```
dariadomagala: Using ace editor is a good idea as long as the lambda returns an object, which might not always be the case. If you still want to use the ace editor, please check if it's possible to hide these errors, it shouldn't be visible for the user since it might confuse him that something is wrong.
Screen Shot 2019-05-24 at 10 08 03

dariadomagala: Also, I would make it a little bit bigger ;)
According to automatix team, the response might be long, and it would be difficult to read it seeing only 4 lines.

dariadomagala: Third thing, I think we should clean the 'response' section after each call. I can still see old response even if my newest request failed and in my opinion it might be confusing if someone won't see the error, and will think that this is the current response.

dariadomagala:
https://github.com/kyma-project/console/pull/867#discussion_r287261102
This comment is still valid

lambda/src/app/lambdas/lambda-details/lambda-details.component.html
parostatkiem: Are you sure? 
The _Response_ is cleared when:
- there's no data to show
- there was an error

I tested both of those cases and it works.

Please provide a case when it's not done but should be.

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -1297,4 +1320,103 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
...
+    if (this.notificationTimeout) {
+      clearTimeout(this.notificationTimeout);
+    }
+
+    setTimeout(() => {
+      this.currentNotification = notificationData;
+    });
+
+    if (timeout) {
+      this.notificationTimeout = setTimeout(() => {
```
dariadomagala: why do we even need this variable?

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -1297,4 +1320,103 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
...
+  }
+
+  showNotification(notificationData: INotificationData, timeout?: number) {
+    this.currentNotification = { message: '', type: 'info' }; // "pretend" there's no error to force notification re-render
+
+    if (this.notificationTimeout) {
+      clearTimeout(this.notificationTimeout);
+    }
+
+    setTimeout(() => {
```
dariadomagala: why timeout?

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
parostatkiem: To reset the Notification timeout every time the new Notification has been displayed. Without it, a user could experience an unexpected notification behaviour while clicking `make a test call` button rapidly ⏩

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
parostatkiem: To force Angular to rerender Notification component which is for triggering the animation.
This timeout just puts setting the new value to the end of browser's task queue.

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
parostatkiem: You can test it by heading to the previous commit and clicking the button rapidly ;)

PR #883, opened 2019-05-28
Expose APIs accessibility

parostatkiem:
**Description**

- Add tooltips to _Service Port_ (in tables and API creation/edit mode)
- move _Port_ column away from _Hostname_ in Exposed APIs table
- Make _Hostname_ clickable and add "Copy to clipboard" button next to it

#### `API_PORT_DESCRIPTION` const is up for discussion 
**Related issue(s)**

### DIFF
--- a/core/src/app/app.module.ts
+++ b/core/src/app/app.module.ts
@@ -117,7 +117,7 @@ import { getMainDefinition } from 'apollo-utilities';
 import { split } from 'apollo-link';
 import { AppConfig } from './app.config';
 import { WebSocketLink } from './ws';
-import GenericHelpersService from './shared/services/generic-helpers.service';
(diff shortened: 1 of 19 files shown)


kwiatekus:
Good work! @parostatkiem. I found few potential improvements:
 - please also add the same link and copy button to the list of exposed APIs on the service details
 - please also add the link and copy button to API details page (inside the header component.. just like the `Documentation` link in the following [example](https://app.zeplin.io/project/5ab0db46508f1e7a678a01c1/screen/5cdab3ddb4655d1e742c42c1))

As for the service ports. I just checked that for the user, the service port value is absolutely irrelevant.
Here is a follow up issue to remove service ports: https://github.com/kyma-project/console/issues/894

Issue #876, opened 2019-05-23
Stay on lambda details view after saving/creating a lambda function


kwiatekus:
As a lambda developer I want to stay on lambda details view after applying changes, so that I can continue testing and improving the code.

AC:
 - User can see success Notification after  (`/shared/components/...`)

PR #884, opened 2019-05-28
Stay on lambda details after creation

y-kkamil:
**Description**

- Shows success notification message after lambda create/update action
**Related issue(s)**
Check #876


y-kkamil:
as this was branched off #867, it might be better to wait for it to be merged.

parostatkiem:
I haven't looked at the code yet but I noticed few things to improve:
- When I create lambda, then click `Create`, I can still **Create** a lambda ;) Please change "create" mode to "edit" (if the code provides such a possibility ;) ;) ;) )
- When I edit or create a lambda which has no HTTPS endpoint and then I add an HTTPS endpoint, I'm still unable to test the lambda. Please update `existingHTTPEndpoint` variable or (seems better for me) re-fetch the whole lambda data (somehow trigger the subscription mechanizm)

kwiatekus:
great job @y-kkamil. I found some potential improvements:
 - create a lambda, give it a name and hit 'create'
- attach https trigger, hit save 
- modify lambda body, hit save --> 409 error
Screen Shot 2019-05-30 at 16 34 44

y-kkamil:
it seems like the solution introduces another bug:
if you have a lambda with an event trigger and you delete it, then after save, more often than not (it's not deterministic), the trigger will come back from the service anyway :( 

therefore, I am closing the PR until we discuss another solution.

PR #885, opened 2019-05-28
Fix errors when lambda's yaml is minimal

parostatkiem:
**Description**

- Display environment variables only if a suitable path of properties exists in YAML
- fix another bug connected with inexistent `metadata.annotations`

Tested with the following minimal yaml:
```yaml
---
apiVersion: kubeless.io/v1beta1
kind: Function
metadata:
  name: hello3
spec:
  function: |-
    module.exports = {
        main: function (event, context) {
            return 'hello world';
        }
    }
  runtime: nodejs8
  handler: handler.main

```

**Related issue(s)**


lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -869,12 +869,15 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
 
           this.setLoaded(true);
           this.initializeEditor();
-          this.functionSizes.forEach(s => {
+          if(lambda.metadata && lambda.metadata.annotations)
```
sjanota: Three lines below `{` is placed just after the if statement, not in a new line. I'd suggest picking one style.

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -1231,7 +1233,22 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
...
+    this.lambda.spec.deployment = {
+      ... this.lambda.spec.deployment,
+      spec:{
+        template:{
+          spec:{
+            containers:[{
+              env: $event,
+              name: this.lambda.metadata.name,
+              image: '',
+              resources:{}
```
y-kkamil: these two lines above will override both `image` and `resources` properties. don't they affect the lambda deployment object then? 
i can see, that after env vars modification the information about memory and cpu for this particular deployment is lost.

parostatkiem: Any suggestion of the solution? I tried to avoid a vast tree of `&&` and conditional property creation 😫

y-kkamil: i'll try to come up with something ;)

y-kkamil:
I'm closing this one in favour of #900

Issue #834, opened 2019-05-13
Test lambda view uses real event examples


sjanota:
@sjanota commented on [Mon May 13 2019](https://github.com/kyma-project/kyma/issues/4033)

**Description**

Allow to test the lambda via https call from the console UI using generated sample payloads that correspond to events available in the namespace.

AC:
 - Once lambda is exposed via https, user can trigger it using sample payload
 - User can select any event (available in the namespace) on the Testing tab and see the generated payload
 - User can edit the payload freely

**Reasons**

Lambda developer should be able to test the lambda behaviour before it is wired to the system's business events

PR #892, opened 2019-05-30
Display event trigger schema example

akucharska:
**Description**

- Once lambda is exposed via https, user can trigger it using sample payload
- User can select any event (available in the namespace) on the Testing tab and see the generated payload
- User can edit the payload freely

**Related issue(s)**
https://github.com/kyma-project/console/issues/834

### DIFF
--- a/lambda/src/app/lambdas/lambda-details/lambda-details.component.scss
+++ b/lambda/src/app/lambdas/lambda-details/lambda-details.component.scss
@@ -145,3 +145,31 @@ input[type='text']:focus {
 .flex-right{
   justify-content: flex-end;
 }
+
+.y-fd-form__control[type='text'] {
+  border: 0;
+  padding-right: 40px;
+
+  &:focus {
+    box-shadow: none;
+  }
+}
+
+.y-fd-dropdown__control {
+  height: auto;
+  max-height: none;
+
+  &:after {
+    position: absolute;
+    top: 0;
+    right: 0;
+    height: 100%;
+    margin: 0;
+  }
+}
+
+.y-fd-popover__body {
+  width: calc(100% - (var(--fd-modal-padding-x)*2));
+  overflow: auto;
+  max-height: 300px;
+}
--- a/lambda/src/app/lambdas/lambda-details/lambda-details.component.scss
+++ b/lambda/src/app/lambdas/lambda-details/lambda-details.component.scss
@@ -169,7 +169,11 @@ input[type='text']:focus {
 }
 
 .y-fd-popover__body {
-  width: calc(100% - (var(--fd-modal-padding-x)*2));
+  min-width: 100%;
   overflow: auto;
   max-height: 300px;
+}
+
+.y-fd-popover__body--right {
+  right: 0;
 }
--- a/lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
+++ b/lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
@@ -1415,8 +1415,8 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
         this.showNotification(
           {
             type: `success`,
-            message: `Test Event Sent. Check Lambda Logs.`,
-            description: `You can now browse its logs.`,
+            message: `Test Event Sent.`,
+            description: `Check Lambda Logs.`,
           },
           5000,
         );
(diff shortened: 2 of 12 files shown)


lambda/src/app/lambdas/lambda-details/lambda-details.component.scss
```diff
@@ -145,3 +145,31 @@ input[type='text']:focus {
...
+    position: absolute;
+    top: 0;
+    right: 0;
+    height: 100%;
+    margin: 0;
+  }
+}
+
+.y-fd-popover__body {
+  width: calc(100% - (var(--fd-modal-padding-x)*2));
```
parostatkiem: Is it somehow connected with modal padding?

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -78,7 +78,10 @@ interface INotificationData {
 })
 export class LambdaDetailsComponent implements OnInit, OnDestroy {
   selectedTriggers: ITrigger[] = [];
+  selectedEventTrigger: ITrigger[] = [];
```
parostatkiem: `selectedEventTriggers`?

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -119,6 +123,9 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
   showSample = false;
   toggleTrigger = false;
   toggleTriggerType = false;
+  trigerDropdownHidden = true;
```
parostatkiem: It's not used anywhere.

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -119,6 +123,9 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
...
+  public triggerAriaExpanded = false;
+  public triggerAriaHidden = true;
```
parostatkiem: Can't it be just one boolean instead of 2?
For me, even better would be to name it `isTriggerDropdownExpanded`.

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -1424,4 +1445,43 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
...
+  filterTriggerEvents() {
+    this.filteredTriggers = [];
+    this.allEventTriggers.forEach(element => {
+      if (element.eventType.toLowerCase().indexOf(this.trigger.toLowerCase())!==-1) {
+        this.filteredTriggers.push(element);
+      }
+    });
+  }
+
+  public toggleDropDown() {
```
parostatkiem: How about merging those 3 similar functions into one (assuming you did the change I suggested above)?
```typescript
 public toggleDropDown(forceState?:boolean) {
    this.isTriggerDropdownExpanded = forceState===undefined ? !this.isTriggerDropdownExpanded : forceState;
  }
 ```

File: lambda/src/app/lambdas/lambda-details/lambda-details.component.ts [hunk 6]
```diff
@@ -1424,4 +1445,43 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
...
+    this.triggerAriaExpanded = true;
+    this.triggerAriaHidden = false;
+  }
+
+  public closeDropDown() {
+    this.triggerAriaExpanded = false;
+    this.triggerAriaHidden = true;
+  }
+
+  private generateExample(schema: JSON) {
```
parostatkiem: If `sample()` always returns an object, please define type of this function.

lambda/src/app/shared/datamodel/event-trigger-with-schema.ts
```diff
@@ -0,0 +1,10 @@
+import { ITrigger } from './trigger';
+
+export class EventTriggerWithSchema implements ITrigger {
```
parostatkiem: I think that could work.
```typescript
import { EventTrigger } from './event-trigger';

export class EventTriggerWithSchema extends EventTrigger implements ITrigger {
  schema: JSON;
}
```

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
akucharska: Not always

PR #900, opened 2019-06-03
Crashing lambda ui

y-kkamil:
**Description**

Reworked solution from PR #885


lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -1231,6 +1233,8 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
   }
 
   handleEnvEmitter($event): void {
+    this.lambda = this.lambdaDetailsService.initializeMissingLambdaContainers(this.lambda);
+
```
parostatkiem: Can't we simply `initializeMissingLambdaContainers` every time new Lambda data is received?
That would let us simplify `getEnvs()` a lot by doing just one property check.

lambda/src/app/lambdas/lambda-details/lambda-env/lambda-env.component.ts
```diff
@@ -6,7 +6,7 @@ import {
   ViewChild,
   ViewContainerRef,
   Output,
-  EventEmitter,
+  EventEmitter
```
parostatkiem: Is it used?

lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
y-kkamil: fixed

y-kkamil:
Another issue, which is somehow connected found: #914

PR #924, opened 2019-06-06
Fetch APIs via gql

parostatkiem:
**Description**

- Use `AbstractGraphqlElementListComponent` for `apis` and `filtered-apis`
- remove all unnecessary code from above components
- add missing types to some functions responsible for rendering

**Related issue(s)**


core/src/app/content/namespaces/configuration/apis/filtered-apis/filtered-apis.component.ts
```diff
@@ -1,107 +1,111 @@
...
+      currentNamespaceService,
+      commService,
+      graphQLClientService,
+      changeDetector
+    );
 
-    const query = `query API($namespace: String!, $serviceName: String!) {
+    this.route.params.subscribe(
+      params => {
+        this.serviceName = params['name'];
```
dariadomagala: We don't need that as a global variable

core/src/app/content/namespaces/configuration/apis/filtered-apis/filtered-apis.component.ts
```diff
@@ -1,107 +1,111 @@
...
+            type
+            issuer
+          }
+        }
+        type
+      }
+    }`;
   }
 
   public getResourceUrl(kind: string, entry: any): string {
```
dariadomagala: Same here, we don't need this function anymore.

core/src/app/content/namespaces/configuration/apis/api-definition-entry-renderer/api-definition-entry-renderer.component.html
```diff
@@ -2,37 +2,36 @@
...
       [size]="'m'"
       title="Copy to clipboard"
     ></fd-icon>
   </td>
-  <td>{{ entry.spec.service.name }}</td>
-
+  <td>{{ entry.service.name }}</td>
   <td>
-    <p>{{ isSecured(entry) }}</p>
+    <p>{{ isSecured(entry) ? 'Yes' : 'No' }}</p>
```
dariadomagala: I know it's not your change, but please make them look the same :) same thing with the header here
Screen Shot 2019-06-19 at 13 49 16
Screen Shot 2019-06-19 at 13 49 23

PR #947, opened 2019-06-11
Log streams only for newest lambda pod

parostatkiem:
**Description**

- by default, show logs only for the **newest** pod of Lambda
- add ability to _also show outdated logs_ (it's shown only if LogUI has been opened with `function` param set)
- properly align labels with text in LogUI
- query all pods from current namespace in LambdaDetails to determine the newest pod for a Lambda

**Related issue(s)**


logging/src/app/search-form/search-form.component.ts
```diff
@@ -95,33 +120,60 @@ export class SearchFormComponent implements OnInit, OnDestroy {
...
+  private fiterStreamByInstance(
+    stream: ILogStream,
+    getLabelValueFn: (label: string) => string,
+  ): boolean {
+    // if a stream has instance="xyz" label, check if "xyz" is in podsForFunction list.
+    const instanceLabel = stream.availableLabels.find((label: string) =>
+      label.includes('instance'),
+    );
+    if (!instanceLabel || !this.podsForFunction) {
+      return true; // this stream doesn't have an 'instance' label
```
y-kkamil: btw why do want to show a stream with no `instance` label? when such a stream even appears in the log?

PR #965, opened 2019-06-19
Add a possibility to hide navigation nodes

dariadomagala:
**Description**

- Add a possibility to hide navigation nodes

**Related issue(s)**

### DIFF
--- a/core/src/luigi-config/luigi-config.js
+++ b/core/src/luigi-config/luigi-config.js
@@ -224,6 +224,8 @@ function getNodes(context) {
     values.forEach(function (val) {
       nodeTree = [].concat.apply(nodeTree, val);
     })
+    
+    hideDisabledNodes(nodeTree, true);
     return nodeTree;
   })
     .catch((err) => {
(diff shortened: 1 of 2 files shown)


core/src/luigi-config/luigi-config.js
```diff
@@ -644,7 +647,10 @@ Promise.all(initPromises)
             }
           ];
           var fetchedNodes = [].concat.apply([], clusterMicrofrontendNodes);
-          return [].concat.apply(staticNodes, fetchedNodes);
+          const nodeTree = [].concat.apply(staticNodes, fetchedNodes);
```
parostatkiem: Why not 
```javascript
const nodeTree=[...staticNodes, ... clusterMicrofrontendNodes];
```
?

core/src/luigi-config/navigation-helpers.js
```diff
@@ -0,0 +1,34 @@
...
+          : 'namespace'
+        : nodeCategory;
+      
+      const nodeLabel = node.label 
+        ? node.label.split(/ |-/).join('').toLowerCase()
+        : '';
+      const nodeId = `${categoryId}${categoryId && nodeLabel ? '.': ''}${nodeLabel}`;
+      
+      const shouldBeDisabled = (element) => {
+        return (element === categoryId || element === nodeId);
```
parostatkiem:
```javascript
         const shouldBeDisabled = element => 
            element === categoryId || element === nodeId;
```

core/src/luigi-config/luigi-config.js
dariadomagala: I didn't change the logic here, just used the result as a variable below. There is a different task to cleanup the luigi-config, so I didn't want to change the old code.

After your review comments, I decided to change that, but your solution wouldn't work anyway since `clusterMicrofrontendNodes` is an array of arrays, so I changed it a bit differently, pleas check it :)

PR #1000, opened 2019-07-01
Organize LogUI toggles

parostatkiem:
**Description**

- Add "Auto refresh" button next to the "Search" button
- Display a tooltip on "Search" button to explain its disabled state when auto refresh is on
- Add new input form item - _"Show" > "Outdated logs"_ only for Lambda logs

### "I can see an **error** when I pause and play auto refresh on freshly opened non-lambda LogUI"
Don't worry ;) There's a bug in the whole search mechanizm - _search query_ field is not validated. When you play auto refresh back (after it was paused), a single form submit is performed (like you also clicked "Search") which makes the backend throw the error because _search query_ is empty.

### DIFF
--- a/logging/src/app/search-form/shared/constants.ts
+++ b/logging/src/app/search-form/shared/constants.ts
@@ -1,2 +1,1 @@
 export const REFRESH_INTERVAL = 2 * 1000; // ms
-export const DISABLED_SEARCH_BUTTON_TOOLTIP= `The results are updated automatically`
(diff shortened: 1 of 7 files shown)


logging/src/app/search-form/shared/constants.ts
```diff
@@ -1 +1,2 @@
-export const REFRESH_INTERVAL = 2 * 1000; // ms
\ No newline at end of file
+export const REFRESH_INTERVAL = 2 * 1000; // ms
+export const DISABLED_SEARCH_BUTTON_TOOLTIP= `The results are updated automatically`
```
dariadomagala: I don't think we need to store this variable in the 'shared' file. This text is valid only for log ui since we don't have autorefresh anywhere else.

parostatkiem: ✅

PR #1016, opened 2019-07-04
Fix render labels in table of add-ons-ui

magicmatatjahu:
**Description**

- Fix render labels in table of add-ons-ui

**Related issue(s)**

### DIFF
--- a/add-ons/src/services/Labels.service.ts
+++ b/add-ons/src/services/Labels.service.ts
@@ -20,14 +20,18 @@ const useLabels = () => {
   const getFiltersLabels = (configs: Configuration[]): FilterLabels => {
     const labels: FilterLabels = {};
     configs.forEach(config => {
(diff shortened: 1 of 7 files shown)


add-ons/src/services/Labels.service.ts
```diff
@@ -20,14 +20,18 @@ const useLabels = () => {
...
-        const label = config.labels[key];
-        if (!labels[key]) {
-          labels[key] = [label];
-        } else {
-          labels[key].push(label);
-        }
-      });
+      if (config.labels) {
+        Object.keys(config.labels).forEach(key => {
+          if (config.labels) {
```
aerfio: you check this 2 lines above

aerfio: TS bugs without it :V ok, it can stay

PR #1047, opened 2019-07-09
Update Fundamental in Core and Lambda

parostatkiem:
**Description**
`<esc>` key handling is enabled by default in the new Fundamental-NGX

- Update _Fundamental-NGX_ and _Fiori Fundamental_ libraries in Core
- Update _Fundamental-NGX_ and _Fiori Fundamental_ libraries in Lambda
- Remove tons of errors after the update
- Adjust some visuals, like modal sizes, popovers etc.
- Disable one unit test in Lambda that was constantly failing for an unknown reason (it'll be rewriten soon anyway)

**Related issue(s)**


parostatkiem:
@dariadomagala 
1,4,5,6,7 ✅ 

2. I believe it is an internal issue of Fundamental NGX. Notice the element _@modal-fade_ mentioned in the error - it's not our code. I tried to solve this issue for some time but from my understanding, it shouldn't appear :) I'm afraid an issue in F-ngx repo will be necessary...
3. I got rid of the initial scrollbar made by a popover placed in a modal. It is something  😊 
However, the only way I could solve the double scrollbars issue after the popover is opened was setting `overflow: visible` to the modal body. That would be a really bad idea and would break some other modals.

dariadomagala:
@parostatkiem number 6 is still valid. 
And we should do something about those dropdowns, maybe not within this task, but eventually.. It will be really annoying for users.
Will you please report a bug to fundamentals?

parostatkiem:
https://github.com/SAP/fundamental-ngx/issues/1035

Issue #1043, opened 2019-07-08
Delete Application


kwiatekus:
**Description**

**AC**
 - User can [delete application](https://github.com/kyma-incubator/compass/blob/master/components/director/pkg/graphql/schema.graphql#L483):
   - from the list 
   - from the details view
   - delete operation needs to be confirmed

Issue #1042, opened 2019-07-08
Update Application


kwiatekus:
**Description**

 - User can [update application](https://github.com/kyma-incubator/compass/blob/master/components/director/pkg/graphql/schema.graphql#L482):
   - name
   - description
   - ~labels~ (https://github.com/kyma-project/console/issues/1130)

Issue #1040, opened 2019-07-08
Create Application


kwiatekus:
**Description**

**AC**
 - User can [create application](https://github.com/kyma-incubator/compass/blob/master/components/director/pkg/graphql/schema.graphql#L481) by defining:
   - name
   - description
   - labels
 - User sees an error message if create operation fails

PR #1051, opened 2019-07-10
Create application details view

### LINKED ISSUE #1041: Application Details
**Description**

**AC**
 
- User can see a [details of application](https://github.com/kyma-incubator/compass/blob/master/components/director/pkg/graphql/schema.graphql#L57) 
    - id
    - name
    - description
    - labels (❗️ assume label value may be an object, not a string)
    - status
    - pageable list of APIs
    - pageable list of event APIs
    - ~pageable list of documents (documentation *.md files)~ out of scope for 1.4

Wawrzyn321:
**Description**

- Create application details view

**Related issue(s)**
Resolves #1041 

Some parts of code are yet to be finished in #1040  , #1042  and #1043


akucharska:
One more thing. When we try to open application that don't exist in this tenant graphql error is displayed. It would be cool to manage that and show some info that application don't exist

compass/src/App.scss
```diff
@@ -36,3 +36,7 @@
 .fd-pagination {
   text-align: center;
 }
+
+header {
```
akucharska: I would rather rely on class than atribute since there might be multiple headers used in different purpose

PR #1071, opened 2019-07-12
Create application modal

akucharska:
**Description**

- Add create application modal

**Related issue(s)**
https://github.com/kyma-project/console/issues/1040


compass/src/components/Applications/CreateApplicationModal/CreateApplicationModal.js
```diff
@@ -0,0 +1,173 @@
...
+      <Mutation mutation={CREATE_APPLICATION_MUTATION} variables={{ in: formData }}>
+        {(createApplication, result) => {
+          const { data, loading, error, called } = result;
+          if (loading) {
+            return <div>LOADING</div>;
+          }
+
+          if (error) {
+            console.log("error", error);
+            LuigiClient.uxManager().showAlert({
```
dariadomagala: Luigi alerts should be used in more 'global' purposes (like resource limit exceeded that should be visible even after changing the route), not instead of a modal :) 
We should create new information modal here.

Edit: Something to discuss, we need to check if it's consistent with fiori guidelines

compass/src/components/Applications/CreateApplicationModal/CreateApplicationModal.js
```diff
@@ -0,0 +1,173 @@
...
+      }
+    });
+  };
+
+  onChangeLabel = (value, ss) => {
+    let formData = this.state.formData;
+    let labelsValidated = true;
+
+    try {
+      formData.labels = JSON.parse(value);
```
dariadomagala: It's not enough - you can create an object 
```
{
  "test":"test"
}
```
which is a valid JSON object, but is not a proper label in compass terms and throws an error `Error: GraphQL error: given value 'string' must be a string array`

compass/src/components/Applications/CreateApplicationModal/CreateApplicationModal.js
```diff
@@ -0,0 +1,173 @@
...
+          if (loading) {
+            return <div>LOADING</div>;
+          }
+
+          if (error) {
+            console.log("error", error);
+            LuigiClient.uxManager().showAlert({
+              text: error.message,
+              type: "error",
+              closeAfter: 2000
```
dariadomagala: If we want to keep those alerts, I don't think we should close them so fast - before I realised there was an error, it was already gone. If there will be a longer error message, users won't have time to read / analyse / copy it.

compass/src/components/Applications/CreateApplicationModal/labelsSchema.js
```diff
@@ -0,0 +1,19 @@
+export const labelsSchema = {
+  title: "Example Schema",
```
y-kkamil: it looks the 'title' property is not required for JSON schema. if we want to keep it here, we should think of a more meaningful one, like 'Application Labels' or something?

instances/src/components/ServiceInstanceDetails/ServiceInstanceBindings/CreateCredentialsModal/CreateCredentialsModal.component.js
```diff
@@ -214,6 +214,7 @@ class CreateCredentialsModal extends React.Component {
         key={serviceInstance.name}
         title={title}
         confirmText="Create"
+        cancelText="Cancel"
```
y-kkamil: is this change from instances needed here?

compass/src/components/Shared/JSONEditor.js
```diff
@@ -0,0 +1,58 @@
...
+    const options = {
+      escapeUnicode: false,
+      history: true,
+      indentation: 2,
+      mode: "code",
+      search: true,
+      sortObjectKeys: false,
+      mainMenuBar: false,
+      onChangeText: this.props.onChangeText,
+      schema: this.props.schema ? this.props.schema : null
```
y-kkamil: i think this can be simplified to just `this.props.schema`

compass/src/components/Applications/CreateApplicationModal/CreateApplicationModal.component.js
```diff
@@ -0,0 +1,294 @@
...
+  };
+
+  getApplicationNameErrorMessage = () => {
+    const { invalidApplicationName, applicationWithNameAlreadyExists, formData } = this.state;
+
+    if (invalidApplicationName) {
+      return this.invalidNameMessage(formData.name);
+    }
+
+    if (applicationWithNameAlreadyExists) {
```
dariadomagala: Backend allows to create multiple applications with the same name, so my question is - which way is the right way? :)

compass/src/components/Applications/CreateApplicationModal/CreateApplicationModal.container.js
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+import { graphql, withApollo, compose } from "react-apollo";
+
+import { CREATE_APPLICATION_MUTATION, CHECK_APPLICATION_EXISTS } from "../gql";
+import { SEND_NOTIFICATION } from "../../../gql";
+
+import CreateApplicationModal from "./CreateApplicationModal.component";
+
+const CreateApplicationContainer = ({ client, ...props }) => {
+  const applicationsExists = () => {
```
dariadomagala: I don't think we should make a call for the list of applications on each input value change. In my opinion we should either allow user to create the app with the same name and show the error if backend won't allow it, or get the list of apps only once at the beginning. 

First, please check with X team, maybe we should allow user create multiple applications with the same name, and then all of that is not needed.

compass/src/components/Applications/CreateApplicationModal/CreateApplicationModal.component.js
akucharska: The application name will be unique in specific tenant so checking id name exist will be needed.

Issue #1088, opened 2019-07-22
Switching namespaces fails sometimes


dariadomagala:
**Description**

Going to lambdas throws an error:
![Screen Shot 2019-07-22 at 11.04.53.png](https://images.zenhubusercontent.com/5b8399bea1576a587faa4a80/bf0cbf9b-0cac-4523-84b8-b88813a9a9eb)
or 401

**Expected result**

**Actual result**

**Steps to reproduce**

**Troubleshooting**

PR #1113, opened 2019-07-29
Fix for lambda ui not preloading

y-kkamil:
**Description**

- Add preload URL processing for 'local development' nodes.

Merge together with: https://github.com/kyma-project/kyma/pull/5034

**Related issue(s)**
See #1088

### DIFF
--- a/core/src/luigi-config/local-development-node-converter.js
+++ b/core/src/luigi-config/local-development-node-converter.js
@@ -34,7 +34,7 @@ export default function processNodeForLocalDevelopment(node, spec, config) {
     node.viewUrl = adjustMicroFrontendUrlForLocalDevelopment(node.viewUrl);
   }
   if (isNodeMicroFrontend && hasNodePreloadUrl) {
-    spec.preloadUrl = adjustMicroFrontendUrlForLocalDevelopment(
+    node.localPreloadUrl = adjustMicroFrontendUrlForLocalDevelopment(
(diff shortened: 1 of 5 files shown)


core/src/luigi-config/microfrontend-converter.js
```diff
@@ -1,4 +1,4 @@
-
+import processNodeForLocalDevelopment from './local-development-node-converter';
```
dariadomagala: Maybe it would be better to call it `local-development-microfrontend-converter`? This name is confusing for me now

core/src/luigi-config/local-development-node-converter.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  if (isNodeClusterMicroFrontend) {
+    node.viewUrl = adjustClusterMicroFrontendUrlForLocalDevelopment(
+      node.viewUrl,
+      clusterMicroFrontendDomainBinding
+    );
+  }
+  if (isNodeClusterMicroFrontend && hasNodePreloadUrl) {
+    spec.preloadUrl = adjustClusterMicroFrontendUrlForLocalDevelopment(
```
dariadomagala: Why do we assign this new value to `spec.preloadUrl`? It's not used anywhere else.

y-kkamil: obvious mistake, fixed

PR #1116, opened 2019-07-30
Api hostname

sjanota:
**Description**

- If Api hostname does not contain domain Apis list adds the configured domain to it

**Related issue(s)**


core/src/app/content/namespaces/configuration/apis/api-definition-entry-renderer/api-definition-entry-renderer.component.ts
```diff
@@ -14,7 +15,12 @@ export class ApiDefinitionEntryRendererComponent
   extends AbstractKubernetesEntryRendererComponent
   implements OnDestroy, OnInit {
   public getHostnameURL = this.genericHelpers.getHostnameURL;
+  public addDomainIfMissing = this.genericHelpers.addDomainIfMissing;
```
dariadomagala: I don't think we should move it to generic helpers since it's not generic at all, we use it only here.

sjanota: Fair enough

core/src/app/content/namespaces/configuration/apis/api-definition-entry-renderer/api-definition-entry-renderer.component.ts
```diff
@@ -14,7 +15,12 @@ export class ApiDefinitionEntryRendererComponent
...
   public disabled = false;
+  public hostname: string = this.addDomainIfMissing(
```
dariadomagala: Also - AFAIK we were supposed to change `hostname` to `URL` or something like that

sjanota: It's up to discussion, but [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.2) states it's called "host". So the "hostname" is better than the "URL" here IMO.

sjanota: According to [Wikipedia](https://en.wikipedia.org/wiki/Hostname#Internet_hostnames) when referencing Internet addresses hostname is FQDN.

dariadomagala: Fine for me ;) @kwiatekus ?

sjanota: @kwiatekus 👍

PR #1138, opened 2019-08-05
Change "hostname" in API to "URL" 

sjanota:
**Description**

- Create url-link component to display links with copy and open
- Add getURL function to build url from parts
- Use those in APIs list, API details and service APIs list 

**Related issue(s)**


core/src/app/shared/services/generic-helpers.service.ts
```diff
@@ -1,7 +1,26 @@
...
 @Injectable()
 export class GenericHelpersService {
-  public getHostnameURL = (hostname: string, domain = ''): string =>
-    `https://${hostname}${domain}`;
+  public getURL = ({
+    schema = 'https',
+    host,
+    domain = AppConfig.domain,
+    port = '',
+    path = '/'
```
dariadomagala: What is it for?

core/src/app/shared/services/generic-helpers.service.ts
```diff
@@ -1,7 +1,26 @@
 import { Injectable } from '@angular/core';
+import { AppConfig } from '../../app.config';
 
 @Injectable()
 export class GenericHelpersService {
-  public getHostnameURL = (hostname: string, domain = ''): string =>
-    `https://${hostname}${domain}`;
+  public getURL = ({
+    schema = 'https',
```
dariadomagala: Maybe `protocol`?

sjanota: URL it's called "schema".

core/src/app/shared/services/generic-helpers.service.ts
sjanota: When calling this method you may provide a path. I'm not using it now, but it was just so easy to add :D

core/src/app/shared/services/generic-helpers.service.ts
dariadomagala: 👍 
It's `scheme` but nvm 😄

Issue #121, opened 2018-09-09
Support fragment navigation in console views with tabs


magicmatatjahu:
**Description**

- Support fragment navigation in `service-catalog` and `instances`, e.g.:
  - after entering the link `.../instances/details/redis-instance#bindings` or `.../instances/details/redis-instance#binding-usages` the specified tab is automatically opened,
  - after entering the link `.../instances/details/redis-instance#(documentations|events|etc)` the specified tab is automatically opened,
  -  ~after entering the link `.../instances/details/redis-instance#delete`, `.../instances/details/redis-instance#create-binding`, `.../service-catalog/details/redis#create`, `.../instances/details/redis-instance#plan-parameters` the specified modal with specific functionality/information is automatically opened.~

Please take into account the following views with tabs
 - Lambda ui 
 - permissions (cluster roles, roles, bindings)
 - resources (resource quotas, limit ranges)

**Reasons**

- improve navigation in `service-catalog` and `instances`.

derberg:
@magicmatatjahu please double check state of console and luigi after 0.6 and if we can go forward with this ticket or not

derberg:
still valid

kwiatekus:
We will handle current tab using luigi parameters.
We will not provide parameter support for linking to a specific operations : for example instance creation or deletion.

PR #1146, opened 2019-08-07
Add tab linking to URL

y-kkamil:
**Description**

- Add ability to 'deep link' to a page with particular tab opened.
(there's an edge case in lambda details view - when you link to the 'testing' tab, and the lambda does not have an http trigger enabled, it will still display the tab. the solution would require more work, so i though i would leave it as it is, having in mind upcoming rework of this view)

**Related issue(s)**
#121


catalog/package.json
```diff
@@ -5,7 +5,7 @@
   "private": true,
   "dependencies": {
     "@kyma-project/generic-documentation": "https://github.com/magicmatatjahu/console/raw/integrate-docs-component-2/components/generic-documentation/kyma-project-generic-documentation-0.1.0.tgz",
-    "@kyma-project/luigi-client": "0.4.9",
+    "@kyma-project/luigi-client": "0.6.0",
```
dariadomagala: Why only package.json? Shouldn't we add package-lock as well here? ;)

y-kkamil: you are totally right! :) i've added changed package-lock file

y-kkamil:
> In Instances and Catalog going to url with preselected tab shows proper tab, but content is always the same, no matter which tabs was selected.

this is now fixed

service-catalog-ui/catalog/src/components/ServiceClassList/ServiceClassList.component.js
```diff
@@ -56,7 +57,14 @@ class ServiceClassList extends React.Component {
     };
   }
 
-  setTabFilter = filterValue => {
+  setTabFilter = currentTabIndex => {
+    let filterValue;
```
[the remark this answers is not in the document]
y-kkamil: i'd go for even simpler solution: 
```
  setTabFilter = currentTabIndex => {
    this.props.setServiceClassesFilter('local', currentTabIndex === 0);
  };
```

PR #1148, opened 2019-08-07
Searching Applications ( Runtimes/ Scenarios) (no longer) breaks the page

Wawrzyn321:
**Description**

- Replace fd-react Search component with fd one.
- Adjust list entities property names

**Related issue(s)**
Fixes [#5194](https://app.zenhub.com/workspaces/kyma---all-repositories-5b6d5985084045741e744dea/issues/kyma-project/kyma/5194)


compass/src/shared/components/GenericList/SearchInput.js
```diff
@@ -0,0 +1,115 @@
...
+        const headers = ['name', 'description'];
+        return headers.map(header => {
+          const entryValue = entry[header];
+          if (entryValue && entryValue.includes(searchQuery)) {
+            return entryValue;
+          }
+          return null;
+        });
+      })
+      .filter(suggestion => suggestion);
```
parostatkiem: If you use `flatMap()` (which is a good idea here IMO), you can just return `[]` instead of returning `null` and filtering.

compass/src/shared/components/GenericList/SearchInput.js
```diff
@@ -0,0 +1,115 @@
...
+    return withDuplicatesRemoved(suggestions);
```
parostatkiem: I think 
```javascript
return Array.from(new Set(suggestions));
```
will do the job instead of _withDuplicatesRemoved_ function 😃

compass/src/shared/components/GenericList/SearchInput.js
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  const searchInputChanged = e => {
+    const searchQuery = e.target.value;
+    handleQueryChange(searchQuery);
```
parostatkiem: It may be my subjective feeling but I think you could directly call `handleQueryChange(e.target.value)` and avoid 5 extra lines.

compass/src/shared/components/GenericList/SearchInput.js
```diff
@@ -0,0 +1,115 @@
...
+
+  const openSearchList = () => {
+    const inputField = searchInputRef.current;
+    if (inputField !== document.activeElement) {
+      inputField.focus();
+    }
+  };
+
+  const setInputFocus = isFocused => () => {
+    setSearchHidden(!isFocused);
```
parostatkiem: `()=>setSearchHidden(xxx)` instead?

compass/src/shared/components/GenericList/SearchInput.js
parostatkiem: Also, please import `import 'core-js/es/array/flat-map';`
because of our old friend IE 😉 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#Browser_compatibility

Wawrzyn321: Unfortunately, the first suggestion won't work, as empty arrays will evaluate to empty strings.

Issue #1144, opened 2019-08-06
Bug in Add Service Instance modal 


kwiatekus:
**Description**

Data required for creating service class instance are populated only once.
If you cancel service class provisioning and go for another try, the form is empty and has no options for `Plan` attribute

**Expected result**

The form should be pre-filled in the same no matter if its shown for the first or consecutive time.

**Actual result**

The form is correctly pre-filled only for the first time.

**Steps to reproduce**

1. Log in to console
2. Navigate to a default namespace
3. Navigate to Catalog node
4. Select service class (I.e test class from [test-add-on](https://github.com/kyma-project/bundles/releases/download/0.6.0/index-testing.yaml))
5. Click `Add` Button
 - you can see a modal with a pre-filled form for creating service instance 
<img width="698" alt="Screen Shot 2019-08-06 at 15 02 11" src="https://user-images.githubusercontent.com/6783567/62542028-458e5000-b85b-11e9-952e-65bad3d0b9f0.png">
Click `Cancel`
Click `Add` button again
 :alert: you can see the form is empty, and there are no options for the `plan` property

<img width="699" alt="Screen Shot 2019-08-06 at 15 04 20" src="https://user-images.githubusercontent.com/6783567/62542167-8ab28200-b85b-11e9-9ee0-9bbaee5de4ba.png">

PR #1161, opened 2019-08-13
Fix instance modal when reopening

y-kkamil:
**Description**

- Create instance modal is now filled with necessary data after reopening it.

**Related issue(s)**
See #1144


service-catalog-ui/catalog/src/components/ServiceClassDetails/CreateInstanceModal/CreateInstanceModal.component.js
```diff
@@ -87,6 +85,10 @@ class CreateInstanceModal extends Component {
         plan: planName,
       },
     });
+  };
+
+  componentDidMount() {
+    this.prepareFormData();
```
dariadomagala: Isn't it enough to add it only in 'onShow'?

y-kkamil: it is, i have removed it

PR #1171, opened 2019-08-19
Improve tab counter appearance in Instances

parostatkiem:
**Description**

- StatusIndicator: Use _Badge_ element built in Fundamental-Ngx instead of custom one
- Make tab status persist for active tab
- ...

**Related issue(s)**


.gitignore
```diff
@@ -18,3 +18,6 @@ yarn-error.log*
 **/lib/**/*
 
 /**/whitesource.config.json
+
+catalog/build
```
akucharska: Add also brokers/build

PR #1173, opened 2019-08-19
Display "show" category whenever "function" label is present

parostatkiem:
**Description**
### "There is an error when the query is erased!"
It's a known behaviour and has nothing to do with this change/issue. It's a problem inside Loki which is still in _beta_ version; it may be fixed in the future.

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/logging/src/app/search-form/search-form.component.ts
+++ b/logging/src/app/search-form/search-form.component.ts
@@ -151,7 +151,7 @@ export class SearchFormComponent implements OnInit, OnDestroy {
     this.isFunctionLabelPresent =
       searchQuery &&
       searchQuery.query &&
-      searchQuery.query.indexOf('{function=') > -1;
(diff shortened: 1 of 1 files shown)


logging/src/app/search-form/search-form.component.ts
```diff
@@ -148,6 +148,11 @@ export class SearchFormComponent implements OnInit, OnDestroy {
   refreshResults() {
     const searchQuery: IPlainLogQuery = this.getSearchQuery();
 
+    this.isFunctionLabelPresent =
+      searchQuery &&
+      searchQuery.query &&
+      searchQuery.query.indexOf('{function=') > -1;
```
dariadomagala: We shouldn't expect that the first character will be `{`. I might want to search function in a specific namespace like: `{namespace="default", function="aaa"}` and this check would fail.

PR #1175, opened 2019-08-20
Sort logs 

dariadomagala:
**Description**

- Sort logs by the last entry timestamp 

**Related issue(s)**

### DIFF
--- a/logging/src/app/search-form/search-form.component.ts
+++ b/logging/src/app/search-form/search-form.component.ts
@@ -366,10 +366,10 @@ export class SearchFormComponent implements OnInit, OnDestroy {
     });
   }
 
-  sortFromNewestLogs(stream1, stream2) {
+  sortFromNewestLogs(stream1: ILogStream, stream2: ILogStream): number {
     if (!(stream1 && stream1.entries && stream2 && stream2.entries)) { return 0 };
 
-    const getTimestamp = (stream) => {
+    const getTimestamp = (stream: ILogStream) => {
       const streamLastEntry = stream && stream.entries && stream.entries.length ? stream.entries[stream.entries.length - 1] : {};
       const streamLastEntryTimestamp =  new Date(streamLastEntry && streamLastEntry.ts ? streamLastEntry.ts : 0);
       return streamLastEntryTimestamp;
--- a/logging/src/app/search-form/search-form.component.ts
+++ b/logging/src/app/search-form/search-form.component.ts
@@ -423,13 +423,13 @@ export class SearchFormComponent implements OnInit, OnDestroy {
   }
 
   public isSearchResultEmpty(searchResult: ISearchResult): boolean {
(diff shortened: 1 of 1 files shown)


logging/src/app/search-form/search-form.component.ts
```diff
@@ -363,6 +366,21 @@ export class SearchFormComponent implements OnInit, OnDestroy {
     });
   }
 
+  sortFromNewestLogs(stream1, stream2) {
```
parostatkiem:
```suggestion
  sortFromNewestLogs(stream1: IlogStream, stream2: IlogStream) {
```

logging/src/app/search-form/search-form.component.ts
```diff
@@ -363,6 +366,21 @@ export class SearchFormComponent implements OnInit, OnDestroy {
...
+    if (!(stream1 && stream1.entries && stream2 && stream2.entries)) { return 0 };
+
+    const getTimestamp = (stream) => {
```
parostatkiem:
```suggestion
    const getTimestamp = (stream: IlogStream) => {
```

logging/src/app/search-form/search-form.component.ts
```diff
@@ -405,12 +423,13 @@ export class SearchFormComponent implements OnInit, OnDestroy {
   }
 
   public isSearchResultEmpty(searchResult: ISearchResult): boolean {
+    const nonEmptyStreams = searchResult.streams.filter(stream => {
```
parostatkiem:
```javascript
return searchResult.streams.some(s: ILogStream => !(s.entries && s.entries.length))
```

dariadomagala: more like 
`return searchResult.streams.every(s: ILogStream => !(s.entries && s.entries.length))`
because `some` would return `true` if there would be **at least one** empty stream in an array of many streams where the rest of them could be non-empty, and that's not what we want.
Also, in my approach I first filtered out all empty streams not to show them like that since it doesn't bring any value:
Screen Shot 2019-08-21 at 11 08 07

parostatkiem: Sorry, I got wrong the function's intention. To achieve such result, small change needs to be made to my solution:
```javascript
return !searchResult.streams.some(s: ILogStream => s.entries && s.entries.length)
```

dariadomagala: Sure, it will be the same result as with `every` so it won't hide all results, but it will still show those empty log streams.

parostatkiem: I believe it will not.

dariadomagala: Ok, my bad.. I did this empty streams filtering thing somewhere else, and somehow didn't commit it.. You're right, we don't need filtering here. 
Although I added one line in other place to filter those empty logs, please check it :)

PR #1189, opened 2019-08-22
Add delete for Metadata Definitions

y-kkamil:
**Description**

- Add delete action to list and details of a Metadata Definition

**Related issue(s)**
See #1129


compass/src/components/Runtimes/RuntimeDetails/RuntimeDetailsHeader/RuntimeDetailsHeader.component.js
```diff
@@ -87,7 +59,15 @@ class RuntimeDetailsHeader extends React.Component {
             </section>
             <ActionBar.Actions>
               <Button
-                onClick={() => this.handleDelete(this.props.runtime)}
+                onClick={() =>
+                  handleDelete(
+                    'Runtime',
+                    this.props.runtime.id,
```
dariadomagala: Just `id` will be enough, similar with `name`. The same thing in applicationDetailsHeader.

compass/src/components/MetadataDefinitions/MetadataDefinitions.component.js
```diff
@@ -21,6 +21,24 @@ class MetadataDefinitions extends React.Component {
...
+      handler: entry => {
+        handleDelete(
+          'Metadata Definition',
+          entry.key,
+          entry.key,
+          this.props.deleteLabelDefinition,
+          this.props.labelDefinitions.refetch,
+        );
+      },
+      skipAction: function(entry) {
```
dariadomagala: I know it's not in the scope of this task, but since you introduced this option here, could you please add something like 
```
skipAction: (() => {
  return !(scenarios && scenarios.length > 1)
})
```
to `ApplicationDetailsScenarios.component.js`? 🙂  The 'unassign' action should not be performed if there is only one scenario assigned.

PR #1195, opened 2019-08-26
Add MetadataDefinitions UI tests

parostatkiem:
**Description**

- Add MetadataDefinitions UI tests
- ...
- ...

**Related issue(s)**


compass/package.json
```diff
@@ -35,11 +35,12 @@
     "xml-js": "^1.6.11"
   },
   "scripts": {
-    "start-live-server": "live-server --entry-file=index.html --mount=/luigi-core:./node_modules/@kyma-project/luigi-core public-luigi/",
     "start": "concurrently \"PORT=8888 react-scripts start\" \"npm run buildConfig:watch\" \"wait-on http://localhost:8888/ && npm run start-live-server\" ",
+    "start-live-server": "live-server --quiet --entry-file=index.html --mount=/luigi-core:./node_modules/@kyma-project/luigi-core public-luigi/",
     "start:kyma": "echo 'This component is not ready yet to be served together with Console.'",
     "build": "react-scripts build && ncp 'node_modules/@kyma-project/luigi-core' 'public-luigi/luigi-core'",
     "test": "react-scripts test",
+    "test-ui": "jest -t='UI'",
```
dariadomagala: Running `npm test` (which also runs ui-tests) works fine, but this script not only also fires all tests but throws errors like:
```
 FAIL  src/components/MetadataDefinitions/test/MetadataDefinitions.ui.test.js
  ● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    SyntaxError: /Users/___/Desktop/Kyma-project/console/compass/src/components/MetadataDefinitions/test/MetadataDefinitions.ui.test.js: Unexpected token (12:6)

      10 |   it(`Renders "loading" when there's no GQL response`, async () => {
      11 |     const component = mount(
    > 12 |       <MockedProvider addTypename={false}>
         |       ^
      13 |         <MetadataDefinitions />
      14 |       </MockedProvider>,
      15 |     );
```
Every test with this script fails.

compass/src/setupTests.js
```diff
@@ -1,10 +1,15 @@
...
 import Adapter from 'enzyme-adapter-react-16';
+import 'jsdom-worker-fix';
 
 var nodeCrypto = require('crypto');
 global.crypto = {
   getRandomValues: function(buffer) {
     return nodeCrypto.randomFillSync(buffer);
   },
 };
+global.URL.createObjectURL = jest.fn();
```
dariadomagala: what is it for?

compass/src/components/MetadataDefinitions/MetadataDefinitionDetails/test/MetadataDefinitionDetails.ui.test.js
```diff
@@ -0,0 +1,160 @@
...
+
+    it(`Renders panel with toggle set to off`, async () => {
+      component.update();
+      expect(
+        component
+          .find(Panel)
+          .find('Toggle')
+          .prop('checked'),
+      ).toBeUndefined();
+      //   expect(console.error.mock.calls[0][0]).toMatchSnapshot(); // unique "key" prop warning
```
dariadomagala: Uncomment it please.

compass/src/setupTests.js
parostatkiem: It's one of pollyfills for JSDom virtual browser.

PR #1217, opened 2019-09-05
Add missing constant in config for local development

parostatkiem:
**Description**

- Fix error that prevented AddOns from rendering when run not on cluster
- ...
- ...

**Related issue(s)**

### DIFF
--- a/config.js
+++ b/config.js
@@ -25,4 +25,12 @@ window.clusterConfig = {
   addOnsModuleUrl: 'http://' + localDomain + ':8004',
   lambdasModuleUrl: 'http://' + localDomain + ':4201',
   logsModuleUrl: 'http://' + localDomain + ':4400',
+
+  HELM_BROKER_REPO_URL_PREFIXES: [
+    'https://',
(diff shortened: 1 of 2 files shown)


config.js
```diff
@@ -25,4 +25,12 @@ window.clusterConfig = {
   addOnsModuleUrl: 'http://' + localDomain + ':8004',
   lambdasModuleUrl: 'http://' + localDomain + ':4201',
   logsModuleUrl: 'http://' + localDomain + ':4400',
+
+  HELM_BROKER_REPO_URL_PREFIXES: [
```
dariadomagala: all variables are camel case, could we also do it here? ;)

parostatkiem: No. We would have to change it deep inside Kyma :(

magicmatatjahu: Please add this table [here](https://github.com/kyma-project/console/blob/master/add-ons/src/constants.ts#L5)

parostatkiem: Why?

PR #1220, opened 2019-09-06
Add OWNERS files

sjanota:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -12,10 +12,12 @@ aliases:
     - y-kkamil
     - dariadomagala
     - akucharska
-    content-reviewers:
+    content-ui-reviewers:
     - magicmatatjahu
     - aerfio
-    - m00g3n
+    content-ui-approvers:
+    - magicmatatjahu
+    - aerfio
     documentation-reviewers:
     - kazydek
     - klaudiagrz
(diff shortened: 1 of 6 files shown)


OWNERS_ALIASES
```diff
@@ -0,0 +1,34 @@
...
+    - dariadomagala
+    - akucharska
+    - sjanota
+    default-approvers:
+    - parostatkiem
+    - kwiatekus
+    - y-kkamil
+    - dariadomagala
+    - akucharska
+    content-reviewers:
```
mmitoraj: For me `content` is quite misleading as it may suggest docu ;) Maybe `peer-reviewers` would be better?

kazydek: IMHO, that somehow suggests doc-related reviews.

OWNERS_ALIASES
```diff
@@ -0,0 +1,34 @@
...
+    - magicmatatjahu
+    - aerfio
+    - m00g3n
```
aerfio: Marcin is on holidays, but I think he'd not like to be included here as he has little to no experience in console

OWNERS_ALIASES
sjanota: `content` is a directory in this repository. That's why we used that name. Do you still think it's better to name it differently?

Issue #1024, opened 2019-07-05
Micro frontends pre-loaded in Console UI


kwiatekus:
**Description**

With [luigi 0.5.0](https://github.com/SAP/luigi/blob/master/CHANGELOG.md#v050-2019-07-04) we can preload remote micro front-end `viewgroups`.

We need to introduce a `/preload` route for all micro front-ends, so that console can use it to bootstrap the corresponding webapp (load all bundles, etc) in the background even before user decides to open the view.
Such a route should do nothing else than bootstrap the microfrontend ui.

It was already done for [lambda-ui](https://github.com/kyma-project/kyma/blob/master/resources/core/charts/kubeless/charts/lambdas-ui/templates/lambdas-microfrontend.yaml#L13)

Add preload urls for:
 - catalog-ui
 - instances-ui
 - brokers-ui
 - log-ui
 - add-ons 

**Reasons**

Preloading `viewgroups` will significantly reduce loading time when switching to remote micro front-ends views

PR #1222, opened 2019-09-09
Add generic component for preloading (Service Catalog UIs)

y-kkamil:
**Description**

- Add preload mechanism for Catalog, Instances and Brokers views

**Related issue(s)**
See https://github.com/kyma-project/console/issues/1024


service-catalog-ui/preloadHandling.js
```diff
@@ -0,0 +1,12 @@
+export default function handlePreload(startAppFn) {
```
magicmatatjahu: If you already have globally available `common` package then maybe it would be a good idea to consolidate the code in whole repo?

components/react/src/commons/preload-handling.js
```diff
@@ -0,0 +1,12 @@
+export function handlePreload(startAppFn) {
```
magicmatatjahu: Please use here implementation of https://github.com/kyma-project/console/blob/master/common/src/utils/preloadingStrategy.ts
and change logic for extracting pathname to your solution.

y-kkamil: you are totally right, i fixed it 👍

service-catalog-ui/catalog/src/index.js
```diff
@@ -4,17 +4,19 @@ import ReactDOM from 'react-dom';
...
 import { createApolloClient } from './store';
-const client = createApolloClient();
 
-(async () => {
+preloadingStrategy(async () => {
   await builder.init();
+  const client = createApolloClient();
   ReactDOM.render(
     <BrowserRouter>
       <ApolloProvider client={client}>
```
magicmatatjahu: `<ApolloProvider client={client}>` should wrap whole app. Please do it for avoid unnecessary rendering app.

service-catalog-ui/instances/src/index.js
```diff
@@ -23,4 +25,4 @@ import { createApolloClient } from './store';
     </BrowserRouter>,
```
magicmatatjahu: `<ApolloProvider client={client}>` should wrap whole app. Please do it for avoid unnecessary rendering app.

service-catalog-ui/brokers/src/index.js
```diff
@@ -13,7 +15,7 @@ import App from './components/App/App.component';
 
 const client = createApolloClient();
 
-(async () => {
+preloadingStrategy(async () => {
   await builder.init();
   ReactDOM.render(
     <BrowserRouter>
```
magicmatatjahu: `<ApolloProvider client={client}>` should wrap whole app. Please do it for avoid unnecessary rendering app.

core/src/luigi-config/microfrontend-converter.js
```diff
@@ -99,11 +99,11 @@ export default function convertToNavigationTree(
...
         } else {
           node.viewGroup = node.navigationContext;
-          // if (spec.preloadUrl) { TODO: UNCOMMENT TO ENABLE PRELOADING (https://github.com/kyma-project/console/issues/1024)
-          //   navigation.viewGroupSettings[node.viewGroup] = {
-          //     preloadUrl: node.localPreloadUrl || spec.preloadUrl
-          //   };
-          // }
+          if (spec.preloadUrl) {
+            navigation.viewGroupSettings[node.viewGroup] = {
+              preloadUrl: node.localPreloadUrl || spec.preloadUrl
```
magicmatatjahu: Please concatenate here `preloadUrl` with `baseUrl` of microfrontend if `preloadUrl` doesn't start as `baseUrl`. As I remember, if I pass to `preloadUrl` eg `preload` then preloading doesn't work, because `green guy` looks only for `preload`, so for future better would be combinating those two values to one in our side.

y-kkamil: according to the Luigi docs [here](https://github.com/SAP/luigi/blob/c17c0801faaa333e0039758de6fef6aa8d7dc0b6/docs/navigation-parameters-reference.md#node-navigation-parameters), preloadUrl should always be an absolute URL, so I don't expect any URL fragments here.

PR #1248, opened 2019-09-11
Clean up after Gulp introduction

parostatkiem:
**Description**

- Remove as many dependencies as possible in Core
-  Remove as many dependencies as possible in Compass
-  Remove as many dependencies as possible in Lambda

**Related issue(s)**


compass/package.json
```diff
@@ -3,23 +3,16 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@kyma-project/luigi-client": "0.4.12",
     "@kyma-project/luigi-core": "0.4.12",
```
dariadomagala: Wa have luigi-core in the root, we could use that.

compass/Dockerfile
```diff
@@ -29,7 +30,6 @@ ENV CI true
 
 # Test & Build
 RUN cd /app/compass && npm run buildConfig
-RUN cd /app/compass && npm test
```
dariadomagala: Why are we not running tests here?

compass/package.json
```diff
@@ -3,23 +3,16 @@
...
-    "styled-components": "^3.4.10",
     "uuid": "^3.3.2",
     "uuid-validate": "0.0.3",
     "xml-js": "^1.6.11"
   },
   "scripts": {
-    "start-live-server": "live-server --quiet --entry-file=index.html --mount=/luigi-core:./node_modules/@kyma-project/luigi-core public-luigi/",
-    "start": "concurrently \"PORT=8888 ../node_modules/.bin/react-scripts start\" \"npm run buildConfig:watch\" \"wait-on http://localhost:8888/ && npm run start-live-server\" ",
+    "start-live-server": "../node_modules/.bin/live-server --quiet --entry-file=index.html --mount=/luigi-core:./node_modules/@kyma-project/luigi-core public-luigi/",
+    "start": "../node_modules/concurrently/bin/concurrently.js \"PORT=8888 ../node_modules/.bin/react-scripts start\" \"npm run buildConfig:watch\" \"../node_modules/.bin/wait-on http://localhost:8888/ && npm run start-live-server\" ",
```
dariadomagala: To be consistent with other commands: 
```suggestion
    "start": "../node_modules/.bin/concurrently.js \"PORT=8888 ../node_modules/.bin/react-scripts start\" \"npm run buildConfig:watch\" \"../node_modules/.bin/wait-on http://localhost:8888/ && npm run start-live-server\" ",
```

compass/Dockerfile
parostatkiem: Why would we? We run tests inside docker image few steps later. I think one test run is enough :)

compass/package.json
parostatkiem: It looks like this can't be done. It just doesn't work.

dariadomagala: Then, there is no need to keep it in the root since only core view uses lugi-core from the root. No need to install it for all of them.

kwiatekus: As far as I understand, both `/core` and `/compass` use luigi-core ..
So maybe its worth to keep it in the root package.json only.

PR #1289, opened 2019-09-26
Logs in modal window for pods and lambdas fixed

kwiatekus:
**Description**

- append `splitView` node param to enforce compact view for logs in modals (pods & lambda list)
- add `Show logs` menu entry  basing on `loki` BM existance
- fix log labels for pods 

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/4711

### DIFF
--- a/core/src/app/content/namespaces/operation/pods/pods.component.ts
+++ b/core/src/app/content/namespaces/operation/pods/pods.component.ts
@@ -8,6 +8,7 @@ import { PodsHeaderRendererComponent } from './pods-header-renderer/pods-header-
 
 import * as luigiClient from '@kyma-project/luigi-client';
 import { GraphQLClientService } from 'shared/services/graphql-client-service';
+import { identifierModuleUrl } from '@angular/compiler';
 
 @Component({
   templateUrl: '../kubernetes-element-list.component.html'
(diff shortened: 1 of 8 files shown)


core/src/app/content/namespaces/operation/pods/pods.component.ts
```diff
@@ -87,7 +87,7 @@ export class PodsComponent extends AbstractGraphqlElementListComponent
     handler.showLogs = (entry: any) => {
       luigiClient
         .linkManager()
-        .withParams({ pod: entry.name, namespace: this.currentNamespaceId })
+        .withParams({ app: entry.labels.app, namespace: this.currentNamespaceId, splitViewMode: 'true' })
```
sjanota: `app` label AFAIK is optional. Shouldn't we rely on `pod`?

kwiatekus: the problem is ... the pod label is no longer there

kwiatekus: ... but still, I have changes the logic here.. 
using app label if present and additionally pod-template-hash label, if present

core/src/app/content/namespaces/operation/pods/pods.component.ts
```diff
@@ -8,6 +8,7 @@ import { PodsHeaderRendererComponent } from './pods-header-renderer/pods-header-
 
 import * as luigiClient from '@kyma-project/luigi-client';
 import { GraphQLClientService } from 'shared/services/graphql-client-service';
+import { identifierModuleUrl } from '@angular/compiler';
```
y-kkamil: what's it for here?

PR #1331, opened 2019-10-15
Create function

sjanota:
**Description**

- Add ability to create a function from lambdas list

**Related issue(s)**


core-ui/src/components/App/App.js
```diff
@@ -12,6 +13,11 @@ export default function App() {
         <Route path="/lambdas" exact component={Lambdas} />
         <Route path="/preload" component={() => null} />
         <Route path="/namespaces" component={NamespaceList} />
+        <Route
+          path="/lambdas/:id"
+          exact
```
sjanota: Instead of `exact` we could move it to the top. `Switch` takes care of the `exact` part.

core-ui/src/components/Lambdas/CreateLambdaForm/CreateLambdaForm.js
```diff
@@ -0,0 +1,99 @@
...
+import { K8sNameField } from '../../CreateNamespaceForm/K8sNameField';
+
+CreateLambdaForm.propTypes = {
+  onChange: PropTypes.func,
+  onCompleted: PropTypes.func,
+  onError: PropTypes.func,
+  formElementRef: PropTypes.shape({ current: PropTypes.any }).isRequired,
+};
+
+function convertLabelsArrayToObject(labelsArray) {
```
dariadomagala: Before solving merge conflicts, there was a function called `createLabelsToDisplay` in LabelsInput component - thanks to that we pass labels as an object to this component and we don't have to worry about the rest. With this solution here, we need to add this `convertLabelsArrayToObject` function everywhere where we use LabelsInput - not very efficient.

core-ui/src/components/Lambdas/CreateLambdaForm/CreateLambdaForm.js
```diff
@@ -0,0 +1,99 @@
...
+  return Object.fromEntries(labelsArray.map(label => label.split('=')));
+}
+
+export default function CreateLambdaForm({
+  onChange,
+  onCompleted,
+  onError,
+  formElementRef,
+}) {
+  const [labels, setLabels] = React.useState([]);
```
dariadomagala: If we would change our approach to use `createLabelsToDisplay` function (from previous comment), labels will we objects.

core-ui/src/components/Lambdas/CreateLambdaForm/CreateLambdaForm.js
dariadomagala: Also, with current solution, you can create multiple labels with the same key (like `a=1`, `a=2`, `a=3`) on the UI side, while in fact only one label will be created (`a=3` in this case) and I think it's confusing. With labels as object user will see that previous label was overwritten.

core-ui/src/components/Lambdas/CreateLambdaForm/CreateLambdaForm.js
```diff
@@ -0,0 +1,99 @@
...
+        size: formValues.size.current.value,
+        runtime: formValues.runtime.current.value,
+        namespace: LuigiClient.getEventData().environmentId,
+        labels: convertLabelsArrayToObject(labels),
+      };
+      await createLambda({
+        variables,
+      });
+      onCompleted(variables.name, `Lambda created succesfully`);
+      LuigiClient.linkManager().navigate(`details/${variables.name}`);
```
dariadomagala: URL has changed.

akucharska: It is still valid

PR #1424, opened 2019-11-24
Name the components in shared folder

Wawrzyn321:
**Description**

- Name the components
- Update tests


parostatkiem:
It's better to fix this by changing the test, not the code

PR #1463, opened 2019-12-09
Fix ModalWithForm behaviour

Wawrzyn321:
**Description**

- Introduce `ControlledModal`

**Related issue(s)**
[Kyma](https://github.com/kyma-project/kyma/pull/6572)
[Compass](https://github.com/kyma-incubator/compass/pull/544)

### DIFF
--- a/compass/src/shared/components/ModalWithForm/ModalWithForm.component.js
+++ b/compass/src/shared/components/ModalWithForm/ModalWithForm.component.js
@@ -1,7 +1,8 @@
 import React, { useState, useRef } from 'react';
 import PropTypes from 'prop-types';
-import { Modal } from './../Modal/Modal';
+import { ControlledModal } from './../Modal/ControlledModal';
 import { Button } from 'fundamental-react/Button';
+import LuigiClient from '@kyma-project/luigi-client';
 
 const ModalWithForm = ({
   performRefetch,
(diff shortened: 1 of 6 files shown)


compass/src/shared/components/ModalWithForm/ModalWithForm.component.js
```diff
@@ -55,34 +67,40 @@ const ModalWithForm = ({
...
+      option={button.option}
+      onClick={() => setOpenStatus(true)}
+    >
       {button.text}
     </Button>
   );
 
   return (
-    <Modal
+    <ControlledModal
```
dariadomagala: I believe we can refactor this component to not use `actions` but `cancelText`, `confirmText` and `onConfirm` instead, so the component would look like:
```
    <Modal
      modalOpeningComponent={modalOpeningComponent}
      className={className}
      title={title}
      cancelText='Cancel'
      confirmText={confirmText}
      onConfirm={onConfirm}
    >
```
As I checked it, it worked fine, didn't break any functionality, but is easier to understand, and this `ControlledModal` is not needed at all.

Wawrzyn321: After brief discussion with other team members, I'm rejecting this request regretfully.

PR #1491, opened 2019-12-27
Api details and list

sjanota:
**Description**

- ...
- ...
- ...

**Related issue(s)**


core-ui/package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "@kyma-project/luigi-core": "^0.6.4",
     "@monaco-editor/react": "^2.0.0",
+    "copy-to-clipboard": "^3.2.0",
```
dariadomagala: We have it here, in compass and in the main package.json. We should remove it from here and from compass and leave it only in a shared package.json.

shared/index.js
```diff
@@ -6,10 +6,14 @@ export { InputWithSuffix } from './components/InputWithSuffix/InputWithSuffix';
 import { CollapsiblePanel } from './components/CollapsiblePanel/CollapsiblePanel';
 import { Tooltip } from './components/Tooltip/Tooltip';
 import { PageHeader } from './components/PageHeader/PageHeader';
+export { Spinner } from './components/Spinner/Spinner';
 
 import { handleDelete } from './components/GenericList/actionHandlers/simpleDelete';
 
 import CustomPropTypes from './typechecking/CustomPropTypes';
+export {
+  default as CopiableText,
```
dariadomagala: why not just export it as `CopiableText` in the CopiableText file?

PR #1497, opened 2019-12-27
Update docs-component styles

aerfio:
**Description**

- Update docs-component styles

**Related issue(s)**

See also https://github.com/kyma-project/kyma/issues/6372


service-catalog-ui/catalog/src/components/App/App.js
```diff
@@ -9,7 +9,7 @@ import ServiceClassDetails from '../ServiceClassDetails/ServiceClassDetails';
 import { backendModuleExists } from '../../commons/helpers';
 
 Modal.MODAL_APP_REF = '#root';
-
+// app
```
magicmatatjahu: This is necessary?

magicmatatjahu:
@aerfio One missing thing. What about compass-ui? As I know, there is also used docs-component.

PR #1499, opened 2019-12-30
Update api rule

dariadomagala:
**Description**

- ...
- ...
- ...

**Related issue(s)**


core-ui/src/components/ApiRules/ApiRuleDetails/ApiRuleDetails.js
```diff
@@ -7,13 +7,15 @@ import AccessStrategy from '../AccessStrategy/AccessStrategy';
...
 import { useDeleteApiRule } from '../useDeleteApiRule';
+import EntryNotFound from 'components/EntryNotFound/EntryNotFound';
 
 const ApiRuleDetails = ({ apiName }) => {
   const { error, loading, data } = useQuery(GET_API_RULE, {
     variables: {
       namespace: LuigiClient.getContext().namespaceId,
       name: apiName,
     },
+    fetchPolicy: 'no-cache',
```
parostatkiem: Shouldn't it be set globally?

sjanota: I think we should avoid it as much as possible. There are places where we rely on subscriptions and they don't make sense if the cache is disabled.

PR #1522, opened 2020-01-16
New application list

parostatkiem:
**Description**

- Display Compass apps in the Console
- Add experimental navigation node
- Use a separate Apollo client for compass-related components (there's one for now)
- add `compassGqlApiUrl` to clusterConfig
- add `actionsStandaloneItems` property to GenericList

**Related issue(s)**

### DIFF
--- a/core-ui/src/components/Applications/ApplicationList/ApplicationList.js
+++ b/core-ui/src/components/Applications/ApplicationList/ApplicationList.js
@@ -1,0 +1,5 @@
+import React from 'react';
+
+export default function ApplicationList() {
+  return 'apps';
+}
--- a/core-ui/src/components/Applications/ApplicationList/ApplicationList.js
+++ b/core-ui/src/components/Applications/ApplicationList/ApplicationList.js
@@ -1,5 +1,48 @@
-import React from 'react';
+import React, { useContext } from 'react';
+import { PageHeader, GenericList, Spinner } from 'react-shared';
+import { GET_APPLICATIONS } from 'gql/queries';
+import { useQuery } from '@apollo/react-hooks';
+import { CompassGqlContext } from 'index';
 
 export default function ApplicationList() {
-  return 'apps';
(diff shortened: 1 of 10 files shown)


core-ui/src/components/Applications/ApplicationList/ApplicationList.js
```diff
@@ -0,0 +1,84 @@
...
+    <span
+      className="link"
+      data-test-id="app-name"
+      //   onClick={() => LuigiClient.linkManager().navigate(`details/${item.name}`)}
+    >
+      {item.name}
+    </span>,
+    item.providerName,
+    <Badge>available</Badge>,
+    <Badge modifier="filled" type="success">
```
akucharska: I think status and connected was supposed to be one column :)

parostatkiem: Not after the recent changes 🌞

core-ui/src/components/Applications/ApplicationList/ApplicationList.js
```diff
@@ -0,0 +1,84 @@
...
+import { UNREGISTER_APPLICATION } from 'gql/mutations';
+import { useQuery, useMutation } from '@apollo/react-hooks';
+import { CompassGqlContext } from 'index';
+import Badge from 'fundamental-react/Badge/Badge';
+import LuigiClient from '@kyma-project/luigi-client';
+
+export default function ApplicationList() {
+  const compassGqlClient = useContext(CompassGqlContext);
+  const { data, error, loading, refetch } = useQuery(GET_APPLICATIONS, {
+    // fetchPolicy: 'no-cache',
```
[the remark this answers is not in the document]
parostatkiem: I agree that the default `cache-first` isn't a good idea but I'd go for `cache-and-network`. There's no sense of using `network-only` in all the places because the cache it updates would be never read.

PR #1525, opened 2020-01-21
Display list of applications from Compass and Kyma

parostatkiem:
**Description**

- Use two separate Apollo clients for Compass and Kyma
- Display application data from Compass and Kyma on ApplicationList
- set Compass as the main data source but also try to get and merge in some data from Kyma
- respect subscription from Kyma (that updates the query result)
- add a new (for backward compatibility) `easyHandleDelete` function to simplify the delete dialog/action
- move `notificationContext` to _Shared_
- add predefined `notifySuccess` and `notifyError` methods to `notificationContext` to make things even easier
- move `isSubscriptionOperator` and `mockLink` to shared as they are commonly used

**Related issue(s)**


shared/typechecking/IsSubscriptionOperation.js
```diff
@@ -0,0 +1,7 @@
+export function isSubscriptionOperation({ query }) {
```
Wawrzyn321: It is already defined in `shared/utils/apollo.js`, is it?

shared/components/GenericList/actionHandlers/simpleDelete.js
```diff
@@ -38,3 +38,40 @@ export function handleDelete(
       });
     });
 }
+
+export function easyHandleDelete(
+  entityType,
+  entityName,
+  deleteRequestFn,
+  deleteRequestParam,
```
Wawrzyn321: Wouldn't be better to allow passing only `deleteRequestFn`, and caller could call just
```
easyHandleDelete(
...
deleteRequestFn: () => someDelete(someParam),
...);
```
?
It would work better in case of multiple `deleteFn` arguments.

shared/components/GenericList/actionHandlers/simpleDelete.js
```diff
@@ -38,3 +38,40 @@ export function handleDelete(
...
+  notificationManager,
+  callback = () => {},
+) {
+  Promise.resolve()
+    .then(async () => {
+      try {
+        const result = await deleteRequestFn(deleteRequestParam);
+        const isSuccess =
+          result.data &&
+          (deleteRequestName ? result.data[deleteRequestName] : true);
```
Wawrzyn321: Maybe I'm nitpicking, but it will fail if `result.data[deleteRequestName]` is correct, but has falsy value.

shared/components/GenericList/actionHandlers/simpleDelete.js
```diff
@@ -38,3 +38,40 @@ export function handleDelete(
...
+        if (isSuccess) {
+          notificationManager.notifySuccess({
+            content: `Application ${entityName} deleted`,
```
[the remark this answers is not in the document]
parostatkiem: Good catch!

shared/components/GenericList/actionHandlers/simpleDelete.js
```diff
@@ -38,3 +38,40 @@ export function handleDelete(
...
+          });
+        } else {
+          throw new Exception();
```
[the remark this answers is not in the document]
parostatkiem: Too much C# 😆

shared/components/GenericList/actionHandlers/simpleDelete.js
parostatkiem: I don't think it's even possible.

parostatkiem: As far as I know, you can't request a type without any of its fields in GQL.
```
mutation unregisterApplication($id: ID!) {
    unregisterApplication(id: $id) 
  }
```
would have thrown an error, I believe.

parostatkiem: image

shared/typechecking/IsSubscriptionOperation.js
parostatkiem: Sure it is! This was the place I was going to put it at first, then forgot to remove it.

Issue #1571, opened 2020-02-21
Deprecate UI for API Controller


kwiatekus:
**Description**

Move the navigation node for `APIs` to a deprecated section. 
Move the navigation node for `API Rules` to the `Configuration` category

**Reasons**

The [old API gateway](https://kyma-project.io/docs/components/api-gateway/#overview-overview) is deprecated

**Attachments**
[Console](https://github.com/kyma-project/console/pull/1572)
[Kyma](https://github.com/kyma-project/kyma/pull/7300)

PR #1572, opened 2020-02-21
Deprecate APIs nav node in favor of API Rules nav node

kwiatekus:
**Description**

- Introduce new category `Deprecated`
- Toggle visibility of deprecated nodes
- Display APIRules on `Services` and link them to API Rules view
- Don't display `nodePort` in APIs

**Note**
`APIRules` query is used here without `serviceName` parameter, as it [fails sometimes](https://github.com/kyma-project/kyma/issues/7988).

**Related issue(s)**
https://github.com/kyma-project/console/issues/1571
[Kyma](https://github.com/kyma-project/kyma/pull/7300)


core/src/app/content/settings/organisation/organisation.component.ts
```diff
@@ -57,15 +60,28 @@ export class OrganisationComponent implements OnInit {
...
-    LuigiClient.sendCustomMessage({ id: 'console.toggleExperimental' });
+    this.refreshLeftNavigation();
   }
 
   private refreshContextSwitcher() {
     window.parent.postMessage({ msg: 'luigi.refresh-context-switcher' }, '*');
   }
+
+  private refreshLeftNavigation() {
+    LuigiClient.sendCustomMessage({ id: 'console.toggleExperimental' });
```
dariadomagala: Now it toggles not only experimental views :) 
And I'm not sure what this line actually works? Shouldn't this reload the left nav?

PR #1575, opened 2020-02-24
Display icon for a service with API packages

parostatkiem:
**Description**

- Display the icon only of an appropriate label is set 
- Use fundamental-react components instead of @kyma-project/generic-components in some places
- minor fixes in README.md (not related but too smal for a separate PR)

**Related issue(s)**

### DIFF
--- a/service-catalog-ui/catalog/src/components/ServiceClassList/Cards/Card.scss
+++ b/service-catalog-ui/catalog/src/components/ServiceClassList/Cards/Card.scss
@@ -4,9 +4,9 @@
       &:not(:last-child) {
         margin-right: 6px;
       }
-      $verticalOffset: 4px;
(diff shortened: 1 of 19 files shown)


service-catalog-ui/catalog/src/components/ServiceClassList/Cards/Card.scss
```diff
@@ -0,0 +1,25 @@
+.card {
+  &__header {
+    .icon {
+      &:not(:last-child) {
+        margin-right: 6px;
+      }
+      $verticalOffset: 4px;
+      background-color: #eeeeef;
+      color: #515559;
```
Wawrzyn321: background-color: var(--fd-color-neutral-2);
      color: var(--fd-color-text-2);

PR #1579, opened 2020-02-26
Lambda triggers purely based on knative triggers

kwiatekus:
**Description**

- kyma subscriptions no longer used by lambda UI

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/6763


lambda/src/app/lambdas/lambda-details/lambda-details.component.ts
```diff
@@ -273,21 +270,20 @@ export class LambdaDetailsComponent implements OnInit, OnDestroy {
...
+            this.triggersService
+              .getTriggersForLambda(this.namespace, this.token, lambdaName)
               .subscribe(resp => {
-                resp.items.forEach(sub => {
+                resp.items.forEach(trig => {
                   const evTrigger: EventTrigger = {
-                    eventType: sub.spec.event_type,
-                    version: sub.spec.event_type_version,
-                    sourceId: sub.spec.source_id,
+                    eventType: trig.spec.filter.attributes.type,
```
dariadomagala: Maybe it would be good to set `trig.spec.filter` as a const and check if all of those objects exist at the beginning? 5 levels where each could be undefined might be tricky ;)

PR #1600, opened 2020-03-09
Multiple docs of one type

parostatkiem:
**Description**

- Render multiple APIs of one type in separate tabs
- ...
- ...

**Related issue(s)**

### DIFF
--- a/components/generic-documentation/src/loader/loader.ts
+++ b/components/generic-documentation/src/loader/loader.ts
@@ -70,7 +70,7 @@ export class DocsLoader {
   }
 
   private sortByURL(f1: File, f2: File): number {
-    return f1.url > f2.url ? 1 : -1;
+    return f1.url.localeCompare(f2.url);
   }
 
   private async setSpecification(types: string[]): Promise<void> {
--- a/components/generic-documentation/src/loader/loader.ts
+++ b/components/generic-documentation/src/loader/loader.ts
@@ -69,6 +69,10 @@ export class DocsLoader {
     }
   }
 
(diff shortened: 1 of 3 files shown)


components/generic-documentation/src/renderers/Group.renderer.tsx
```diff
@@ -60,9 +63,10 @@ export const GroupRenderer: React.FunctionComponent<GroupRendererProps> = ({
     luigiClient.getNodeParams().selectedTab || '';
 
   const markdownsExists = existFiles(sources, markdownTypes);
```
dariadomagala: What about markdown?

components/generic-documentation/src/loader/loader.ts
```diff
@@ -205,12 +210,17 @@ export class DocsLoader {
...
-  private extractSpecification(types: string[]) {
+  private extractSpecification(types: string[]): File[] {
     const assets = this.extractAssets(types);
 
-    const file = assets && assets[0] && assets[0].files && assets[0].files[0];
+    const files =
+      assets &&
+      assets
+        .map(asset => asset.files && asset.files[0])
+        .filter(a => a)
```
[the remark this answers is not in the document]
parostatkiem: To filter out assets that has either no `files` field or its first file is falsy. Just in case.

components/generic-documentation/src/renderers/Group.renderer.tsx
parostatkiem: Markdown is treated differentyl. As far as I know, multiple markdowns are somehow displayed in one tab.

PR #1603, opened 2020-03-10
Do not store profile information in web storage

kwiatekus:
**Description**

- profile data not stored in session storage
- groups information parsed from token
- customer admin group renamed to `runtimeNamespaceAdmin`

### DIFF
--- a/core/src/luigi-config/navigation/permissions.js
+++ b/core/src/luigi-config/navigation/permissions.js
@@ -41,7 +41,7 @@ function checkRequiredBackendModules(nodeToCheckPermissionsFor) {
 function isVisibleForCurrentGroup(node) {
   let currentGroups = node.context ? node.context.groups : null;
   if (!Array.isArray(currentGroups)) return true;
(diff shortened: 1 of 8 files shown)


core/src/luigi-config/navigation/permissions.js
```diff
@@ -41,7 +41,7 @@ function checkRequiredBackendModules(nodeToCheckPermissionsFor) {
 function isVisibleForCurrentGroup(node) {
   let currentGroups = node.context ? node.context.groups : null;
   if (!Array.isArray(currentGroups)) return true;
-  const isAdmin = currentGroups.includes(config.adminsGroupName);
+  const isAdmin = currentGroups.includes(config.namespaceAdminGroupName) || currentGroups.includes(config.runtimeAdminGroupName);
```
dariadomagala: Wouldn't it allow the Namespace Admin to see views that should be visible only for the Runtime Admin?

dariadomagala: As discussed: In our use cases those admins should be treated equally

PR #1607, opened 2020-03-11
Don't refresh view on every context change

Wawrzyn321:
**Description**

- Introduce list of observed properties to `ApplicationContextProvider`, so that view is not refreshed unnecessarily (ex. when `parentNavigationContexts` changes).

**Related issue(s)**
https://github.com/kyma-project/console/issues/1608


kwiatekus:
@Wawrzyn321 I noticed that on scenario details view there are duplicated calls to fetch runtimes and scenarios 
Screen Shot 2020-03-13 at 08 14 20

PR #1611, opened 2020-03-12
Move the auth data to the package. instead of a specific api

dariadomagala:
**Description**

- Move the auth data to the package. instead of a specific API

**Related issue(s)**


compass/src/components/ApiPackages/ApiPackageDetails/ApiPackageDetails.js
```diff
@@ -24,6 +25,9 @@ export default function ApiPackageDetails({ applicationId, apiPackageId }) {
   if (error) return <p>`Error! ${error.message}`</p>;
 
   const application = data.application;
+  if (!application)
+    return <p>`Error! Application with ID ${applicationId} not found.`</p>;
```
Wawrzyn321: We can use `ResourceNotFound` here.

shared/index.js
```diff
@@ -1,11 +1,13 @@
...
-import { K8sNameInput } from './components/K8sNameInput/K8sNameInput';
+export { GenericList } from './components/GenericList/GenericList';
+export { K8sNameInput } from './components/K8sNameInput/K8sNameInput';
 export { StringInput } from './components/StringInput/StringInput';
 export { InputWithSuffix } from './components/InputWithSuffix/InputWithSuffix';
 export { InputWithPrefix } from './components/InputWithPrefix/InputWithPrefix';
-import { CollapsiblePanel } from './components/CollapsiblePanel/CollapsiblePanel';
-import { Tooltip } from './components/Tooltip/Tooltip';
-import { PageHeader } from './components/PageHeader/PageHeader';
+export {
```
Wawrzyn321: Could we also group these guys into one barrel file, as in `CredentialForms/index.js`?

PR #1619, opened 2020-03-13
Fix links to applications from namespace details

kwiatekus:
**Description**

- Fetch compassMetadata for applications CR
- Link to appropriate route depending on `enableAPIPackages` flag

### DIFF
--- a/core/src/app/content/namespaces/namespace-details/namespace-details.component.ts
+++ b/core/src/app/content/namespaces/namespace-details/namespace-details.component.ts
@@ -175,7 +175,7 @@ export class NamespaceDetailsComponent implements OnInit, OnDestroy {
     return !application.compassMetadata || !application.compassMetadata.applicationId || application.compassMetadata.applicationId=== '';
   }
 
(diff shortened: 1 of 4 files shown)


kwiatekus:
Do not merge until https://github.com/kyma-project/kyma/pull/7608 is merged

core/src/app/content/namespaces/namespace-details/namespace-details.component.ts
```diff
@@ -166,12 +166,15 @@ export class NamespaceDetailsComponent implements OnInit, OnDestroy {
       .navigate('services');
   }
 
-  public navigateToApplications(applicationName = null) {
-    LuigiClient.linkManager().navigate(
-      applicationName
-        ? '/home/cmf-apps/details/' + applicationName
-        : '/home/cmf-apps'
-    );
+  public navigateToApplications(application = null) {
```
parostatkiem: I think you can remove the default `null` value. The function would run exactly the same way if it was `undefined` and it'll be `undefined` by default without an explicit declaration of it.

PR #1630, opened 2020-03-17
Fix 'docs per plan of null' error

dariadomagala:
**Description**

- Fix docs per plan of null error
- Use a global variable for the docs-per-plan label name

**Related issue(s)**

### DIFF
--- a/service-catalog-ui/catalog/src/components/ServiceClassDetails/ServiceClassDetails.js
+++ b/service-catalog-ui/catalog/src/components/ServiceClassDetails/ServiceClassDetails.js
@@ -124,7 +124,10 @@ export default function ServiceClassDetails({ name, plan }) {
     setCurrentPlan(plans.find(p => p.name === e.target.value));
   }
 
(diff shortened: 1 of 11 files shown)


service-catalog-ui/catalog/src/components/ServiceClassDetails/ServiceClassDetails.js
```diff
@@ -124,7 +124,10 @@ export default function ServiceClassDetails({ name, plan }) {
     setCurrentPlan(plans.find(p => p.name === e.target.value));
   }
 
-  const isAPIpackage = labels[DOCUMENTATION_PER_PLAN_LABEL] === 'true';
+  const isAPIpackage =
+    labels &&
+    labels[DOCUMENTATION_PER_PLAN_LABEL] &&
```
Wawrzyn321: `labels[DOCUMENTATION_PER_PLAN_LABEL] &&` condition is not needed here, the same goes for `Card.component` and `ServiceInstanceTabs.component`.

PR #1634, opened 2020-03-17
Fix namespaces filter dropdown width

dariadomagala:
**Description**

Couldn't recreate such behavior, so at least for now I fixed the width problem.

- Fix namespaces filter dropdown width

**Related issue(s)**


core-ui/src/components/NamespaceList/NamespacesListHeader/NamespaceFilters/NamespaceFilters.js
```diff
@@ -52,8 +52,7 @@ export default function NamespaceFilters({ filters, updateFilters }) {
     <Popover
       body={popoverBody}
       control={popoverControl}
-      placement="bottom-end"
-      widthSizingType="matchTarget"
+      placement="bottom"
       noArrow
```
parostatkiem: I vote for removing this flag. Now, it may not be obvious that the filter button is a dropdown trigger. Here's how it looks without the flag:
image

PR #1636, opened 2020-03-18
Sort tenants by names in tenant chooser

sjanota:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/compass/src/config/luigi-config/helpers/navigation-helpers.test.js
+++ b/compass/src/config/luigi-config/helpers/navigation-helpers.test.js
@@ -15,4 +15,13 @@ describe('getTenantNames', () => {
     expect(options[1].label).toEqual(tenants[0].name);
     expect(options[2].label).toEqual(tenants[2].name);
   });
+
+  it('routes to tenant home page by default', () => {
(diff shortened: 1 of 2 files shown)


compass/src/config/luigi-config/helpers/navigation-helpers.test.js
```diff
@@ -0,0 +1,27 @@
...
+  it('sorts tenants by name', () => {
+    const tenants = [
+      { name: 'B', id: '1234' },
+      { name: 'A', id: '4321' },
+      { name: 'C', id: '6666' },
+    ];
+
+    const options = getTenantNames(tenants);
+
+    expect(options).toHaveLength(3);
```
parostatkiem: A snapshot would be perfect here IMO 👌

parostatkiem: I mean for the whole test.

sjanota: Actually I think the test is even too verbose as it checks `pathValue`, which is not part of the scenario.

PR #1641, opened 2020-03-19
Introduce tenant search modal

Wawrzyn321:
**Description**

- Add possibility to search tenants by name or id
- Distinguish tenants with the same name in the context switcher

**Related issue(s)**

### DIFF
--- a/compass/src/App.component.js
+++ b/compass/src/App.component.js
@@ -16,6 +16,7 @@ import MetadataDefinitions from './components/MetadataDefinitions/MetadataDefini
 import MetadataDefinitionDetails from './components/MetadataDefinitions/MetadataDefinitionDetails/MetadataDefinitionDetails.container';
 import ApiDetails from './components/Api/ApiDetails/ApiDetails';
 import ApiPackageDetails from 'components/ApiPackages/ApiPackageDetails/ApiPackageDetails';
+import TenantSearch from 'components/TenantSearch/TenantSearch';
 
 const NOTIFICATION_VISIBILITY_TIME = 5000;
 
@@ -53,7 +54,15 @@ class App extends React.Component {
         <Notification {...notification} onClick={this.clearNotification} />
         <Router>
           <Switch>
+            <Route
+              path="/preload"
+              component={() => {
+                console.log('load?');
+                return null;
+              }}
+            />
             <Route path="/" exact component={Overview} />
+            <Route path="/tenant-search" exact component={TenantSearch} />
             <Route path="/runtimes" exact component={Runtimes} />
             <Route
               path="/runtime/:id"
--- a/compass/src/App.component.js
+++ b/compass/src/App.component.js
@@ -57,7 +57,7 @@ class App extends React.Component {
             <Route
               path="/preload"
               component={() => {
-                console.log('load?');
+                console.log('load? the same log shows up in core-ui');
                 return null;
               }}
             />
--- a/compass/src/App.component.js
+++ b/compass/src/App.component.js
@@ -54,13 +54,6 @@ class App extends React.Component {
         <Notification {...notification} onClick={this.clearNotification} />
         <Router>
           <Switch>
-            <Route
-              path="/preload"
-              component={() => {
-                console.log('load? the same log shows up in core-ui');
-                return null;
-              }}
-            />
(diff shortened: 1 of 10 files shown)


compass/src/App.component.js
```diff
@@ -53,7 +54,15 @@ class App extends React.Component {
         <Notification {...notification} onClick={this.clearNotification} />
         <Router>
           <Switch>
+            <Route
+              path="/preload"
```
dariadomagala: Did you talk to the Luigi guys? Do we actually need preload here? Especially, that in the current situation we know it doesn't work properly.

compass/src/components/TenantSearch/fetchTenants.js
```diff
@@ -0,0 +1,29 @@
...
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${token}`,
+      },
+      body: JSON.stringify(query),
+    });
+    const json = await response.json();
+    return json.data.tenants;
+  } catch (err) {
+    console.error('Tenants could not be loaded', err);
+    return [];
```
dariadomagala: We should show the error to the user as well, just an empty list might be confusing.

compass/src/config/luigi-config/helpers/navigation-helpers.js
```diff
@@ -80,3 +82,18 @@ export const getTenantNames = tenants => {
   });
   return tenantNames.sort((a, b) => a.label.localeCompare(b.label));
 };
+
+export const customOptionsRenderer = opt => {
+  // we have to manually find selected tenant by id, as Luigi distinguishes
+  // "options" by their labels (which may not be unique)
```
dariadomagala: Wasn't part of this task to also make them unique? I'm not sure, just asking.

compass/src/components/TenantSearch/TenantSearch.js
```diff
@@ -0,0 +1,69 @@
...
+        <span className="link">{tenant.name}</span>
+        <span className="fd-has-color-text-3">{tenant.id}</span>
+      </ListGroup.Item>
+    ))}
+  </ListGroup>
+);
+
+export default function TenantSearch() {
+  const [filter, setFilter] = React.useState('');
+  const { parentPath, token } = LuigiClient.getNodeParams();
```
dariadomagala: can't we just pass it in context instead of nodeparams? We already take tenants from there

compass/src/config/luigi-config/navigation.js
```diff
@@ -20,6 +22,13 @@ const getTenantName = tenantId => {
   return match ? match.name : null;
 };
 
+const openTenantSearch = () => {
+  LuigiClient.setTargetOrigin(window.origin);
+  LuigiClient.linkManager()
+    .withParams({ parentPath: encodeURIComponent(window.location.pathname) })
```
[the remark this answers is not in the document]
Wawrzyn321: To navigate to another tenant while preserving the context (e.g. runtimes, scenarios), we need to know the current address. Inside modal, our pathname is always `/tenant-search`, so we need to pass parent's path when opening the modal.
image

compass/src/components/TenantSearch/TenantSearch.js
Wawrzyn321: `window.referrer` will do here! And token was already passed in context as `idToken`, fixed.

compass/src/config/luigi-config/helpers/navigation-helpers.js
Wawrzyn321: My first solution was to add 4 characters from id, but after a discussion with @kwiatekus we considered it may be confusing for users.

compass/src/components/TenantSearch/TenantSearch.js
```diff
@@ -0,0 +1,85 @@
...
+    const path = getAlternativePath(tenant.id, parentPath);
+    LuigiClient.linkManager().navigate(`/tenant/${path || tenant.id}`);
+  };
+
+  const getFilteredTenants = () => {
+    if (!filter.trim()) {
+      return tenants;
+    }
+    return tenants.filter(
+      tenant => tenant.name.includes(filter) || tenant.id.includes(filter),
```
dariadomagala: Please make it not case sensitive

PR #1643, opened 2020-03-20
luigi rc1 meets console

JohannesDoberer:


Wawrzyn321:
Hey! I've started reviewing your work, and I've noticed an issue - when you try to sign out, the token from storage is removed, but there is no redirection to logout page.
image

PR #1645, opened 2020-03-23
Add the first test using Testcafe

dariadomagala:
**Description**

- Add the first test using Testcafe

**Related issue(s)**


tests/tests/tests.js
```diff
@@ -0,0 +1,75 @@
...
+test('Namespace `default` card is on the Namespaces list', async t => {
+  const iframe = await getIframe();
+  await t
+    .useRole(adminUser)
+    .switchToIframe(iframe)
+    .expect(Selector('.fd-panel__title').withText('default').exists)
+    .ok();
+});
+
+testIfBackendModuleExists(
```
sjanota: Maybe, instead of `if` we should just deploy it as separate testdefinition along with application-connector?

sjanota: Reasoning: backend module is just an implementation detail. What we really want to check is if application-connector is installed it's MF should be available.

tests/config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  login: process.env.LOGIN || 'admin@kyma.cx',
+  password: process.env.PASSWORD || 'nimda123',
```
sjanota: The password is no longer `nimda123` by default, so I'd remove the default value

tests/helpers.js
```diff
@@ -0,0 +1,54 @@
...
+      .typeText('#password', config.password)
+      .click('#submit-login')
+      .wait(5000);
+  },
+  { preserveUrl: true },
+);
+
+const chooseLoginRole = async t => {
+  try {
+    await Selector('#login').visible; // '.exists' doesn't really wait for the selector..
```
sjanota: If it is run on a cluster with multiple dex connectors how long it fails for the selector to fail?

dariadomagala: 5 seconds I believe

tests/Dockerfile
```diff
@@ -1,30 +1,10 @@
...
+COPY package*.json ./
 RUN npm install
-
 LABEL source git@github.com:kyma-project/console.git
-
-COPY /ui-tests /app/src
-
-CMD npm run test:cluster
+COPY . .
+RUN apk add curl
```
Wawrzyn321, line 10: Shouldn't this line be after WORKDIR statement?

sjanota: right, it will speed up things a little probably as this layer won't change that much

tests/README.md
```diff
@@ -2,50 +2,34 @@
...
 ## Development
 
-Learn how to run all tests or specific ones.
-
-### Run all tests
-
-> **NOTE:** You need a valid `kubeconfig` issued for a user with admin rights for the cluster. On local deployments, the appropriate `kubeconfig` file is set automatically during Minikube setup. For cluster deployments, you must obtain the `kubeconfig` file manually.
-
-To run all acceptance tests inside a container in the cluster, use this command:
+To run UI tests inside a container in the cluster using Chromium browser, use this command:
```
alexandra-simeonova, line 13:
```suggestion
To run UI tests inside a container in the cluster using a Chromium browser, use this command:
```

tests/README.md
```diff
@@ -2,50 +2,34 @@
...
 
 ``` bash
 npm run test:cluster
 ```
 
-To run all acceptance tests outside the cluster, use this command:
+To run UI tests inside a container in the cluster using Firefox browser, use this command:
```
alexandra-simeonova, line 19:
```suggestion
To run UI tests inside a container in the cluster using a Firefox browser, use this command:
```

tests/README.md
```diff
@@ -2,50 +2,34 @@
...
 
-To run all acceptance tests during development against `http://console-dev.kyma.local:4200`, use this command:
+To run UI tests during development against `http://console-dev.kyma.local:4200`, use this command:
 
 ``` bash
 npm run test
 ```
 
-> **NOTE:** In the second case, take note of which **GraphQL API** endpoint you use.
+To run UI tests during development against `http://console-dev.kyma.local:4200` using headless browser, use this command:
```
alexandra-simeonova, line 31:
```suggestion
To run UI tests during development against `http://console-dev.kyma.local:4200` using a headless browser, use this command:
```

PR #1652, opened 2020-03-24
Fix searchbar in instances

akucharska:
**Description**

- Fix searchbar in instances

**Related issue(s)**

### DIFF
--- a/service-catalog-ui/instances/src/components/ServiceInstancesList/ServiceInstancesToolbar/styled.js
+++ b/service-catalog-ui/instances/src/components/ServiceInstancesList/ServiceInstancesToolbar/styled.js
@@ -68,4 +68,14 @@ export const Panel = styled(UnstyledPanel)`
 export const SearchWrapper = styled.div`
   max-width: 640px;
   flex: 1 0 auto;
+  .fd-input {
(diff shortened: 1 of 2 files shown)


service-catalog-ui/instances/src/components/ServiceInstancesList/ServiceInstancesToolbar/styled.js
```diff
@@ -68,4 +68,14 @@ export const Panel = styled(UnstyledPanel)`
...
+    background-color: var(--fd-forms-background-color);
+    border-color: var(--fd-forms-border-color);
+    border-radius: 4px;
+    color: var(--fd-color-text-1);
+    &:focus {
+      box-shadow: 0 0 0 1px var(--fd-color-action-focus);
+      --fd-forms-border-color: var(--fd-color-action-focus);
+    }
+  }
 `;
```
Wawrzyn321: The blue background is gone, but the searchbar have different width on dev:
Dev: 
image
Non-dev:
image
Would you kindly see into this? As far as I can see, adding `width: 100%` to `.fd-search-input` should do the trick.

PR #1675, opened 2020-04-02
Common API documentation tab with a chooser

parostatkiem:
**Description**

- change markdownTypes, asyncApiTypes etc. to markdownDefinition object
- introduce `ApiSelector` component and use it in a common API dedicated tab
- place some comments to make `displayName` property introduction easier
- add placeholders for "remembering" the chosen API as a path param (can't be finished without `displayName`)
- **temporarily** use API type as an identifier (the search might feel weird due to this)

**Related issue(s)**


components/generic-documentation/src/renderers/Group.renderer.tsx
```diff
@@ -34,20 +28,44 @@ export enum TabsLabels {
...
     return null;
   }
 
+  const nonMarkdownSources = sources.filter(
+    (s: Source) => !markdownDefinition.possibleTypes.includes(s.type),
+  );
+
+  const apiTabHeader = (
+    <ApiTabHeader>
+      <span>Selected an API to display:</span>
```
akucharska, line 60: **Displayed API** or **Select an API to display** might be better

components/generic-documentation/src/renderers/Group.renderer.tsx
```diff
@@ -34,20 +25,53 @@ export enum TabsLabels {
...
 }) => {
+  const [currentApi, setCurrentApi] = currentApiState;
+  useEffect(() => {
+    if (!currentApi && sources.length && sources[0].type !== 'mock') {
+      // a "mock" source is loaded at first, before the real data arrives
+      setCurrentApi(sources[0]);
+    }
+  }, [currentApi, sources]);
+
+  // useEffect(() => {
```
akucharska, line 45: This is waiting for dispayNames?

parostatkiem: Yes, forgot to add comment.

PR #1684, opened 2020-04-07
Infinite scroll on runtimes

sjanota:
**Description**

- [x] Add infinite scroll on runtimes view in Compass
- [x] Keep the list format as before

**Related issue(s)**


compass/src/components/Runtimes/InfiniteList.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  return (
+    <>
+      <table className="fd-table">
+        <thead className="fd-table__header">
+          <tr className="fd-table__row">
+            {headerRenderer().map((h, index) => (
+              <th className="fd-table__cell" scope="col" key={h || index}>
```
dariadomagala, line 55: please make it a link

compass/src/components/Runtimes/InfiniteList.js
```diff
@@ -0,0 +1,106 @@
+import React, { useState, useEffect } from 'react';
+import { useQuery } from 'react-apollo';
+import PropTypes from 'prop-types';
+
+const InfiniteList = ({
```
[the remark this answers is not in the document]
sjanota: Tow reasons why we didn't:
- we don't have any additional use case so it may be premature 
- in general we don't want to encourage infinite scrolls in console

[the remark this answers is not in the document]
parostatkiem: why?
I think we all agreed an infinite scroll is a really bad way of presenting tabular data. We have only tabular data in Console 🤣

compass/src/components/Runtimes/InfiniteList.js
parostatkiem: Ymmm.. why?  It's header.

sjanota: done

dariadomagala: @parostatkiem wrote it in the wrong place, it was meant to be in a rowrenderer :)

Issue #1618, opened 2020-03-13
Event API Documentation rendering bug


Wawrzyn321:
**Actual result**

![image](https://user-images.githubusercontent.com/10504661/76632748-fb352600-6543-11ea-896c-7acb42f15df3.png)

Spec: [sap-commerce-cloud.txt](https://github.com/kyma-project/console/files/4330160/sap-commerce-cloud.txt)

magicmatatjahu:
AC:
- contribute to compass with conversion from 1.X.X of AsyncAPI spec to 2.X.X by https://github.com/asyncapi/converter-go

michal-hudy:
The specification must be converted to AsyncAPI v2 before it is displayed.

In Kyma the conversion is made by webhook connected to Rafter (it is using https://github.com/asyncapi/converter-go).

PR #1686, opened 2020-04-08
Fix rendering of async api v1 documentation in compass

kwiatekus:
**Description**

- Use [async api js converter](https://github.com/asyncapi/converter-js) against v1 specs before rendering

**Related issue(s)**
https://github.com/kyma-project/console/issues/1618

### DIFF
--- a/compass/src/components/Api/ApiDetails/ApiDetails.js
+++ b/compass/src/components/Api/ApiDetails/ApiDetails.js
@@ -17,6 +17,7 @@ import {
   GET_EVENT_DEFINITION,
 } from './../gql';
 import './ApiDetails.scss';
+import { convert } from 'asyncapi-converter';
 
 const ApiDetails = ({ apiId, eventApiId, applicationId, apiPackageId }) => {
   const queryApi = useQuery(GET_API_DEFININTION, {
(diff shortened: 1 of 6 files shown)


compass/src/components/Api/ApiDetails/ApiDetails.js
```diff
@@ -86,6 +87,19 @@ const ApiDetails = ({ apiId, eventApiId, applicationId, apiPackageId }) => {
...
+  let specToShow = api.spec.data;
+
+  if (eventApiId) {
+    try {
+      const parsedSpec = JSON.parse(specToShow);
+      if (parsedSpec.asyncapi && parsedSpec.asyncapi.startsWith('1.')) {
+        specToShow = convert(specToShow, '2.0.0');
+      }
+    } catch (e) {
+      console.error('Error parsing async api spec', e);
```
parostatkiem, line 100: So, in case the conversion failed we still want to render the spec at 1.x despite knowing it's not 1.x ?

kwiatekus: In case conversion fails, it would still try to render `1.x` version.
Which will end up showing just the top attributes w/o the details.
Like this:
Screen Shot 2020-04-08 at 13 32 56

parostatkiem: Ok, I thought it'll throw an error.

PR #1703, opened 2020-04-16
Allow connection to remote compass-gateway APIs

kwiatekus:
**Description**

- Modify CSP policy for core-ui and compass ui to allow connections to compass-gateway APIs

### DIFF
--- a/core-ui/nginx.conf
+++ b/core-ui/nginx.conf
@@ -43,7 +43,7 @@ http {
         add_header Strict-Transport-Security 'max-age=31536000; includeSubDomains';
         add_header X-XSS-Protection '1; mode=block';
         add_header X-Frame-Options 'SAMEORIGIN';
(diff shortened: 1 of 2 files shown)


core-ui/nginx.conf
```diff
@@ -43,7 +43,7 @@ http {
         add_header Strict-Transport-Security 'max-age=31536000; includeSubDomains';
         add_header X-XSS-Protection '1; mode=block';
         add_header X-Frame-Options 'SAMEORIGIN';
-        add_header Content-Security-Policy "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' https://cdn.jsdelivr.net; style-src 'self' 'unsafe-inline' https://cdn.jsdelivr.net; connect-src 'self' https://*.$kymadomain wss://*.$kymadomain; font-src 'self' data:; frame-ancestors https://*.$kymadomain; object-src 'none'; media-src 'self'; form-action 'self'; img-src * data:; child-src * blob:; worker-src 'self' blob: data:;";
+        add_header Content-Security-Policy "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' https://cdn.jsdelivr.net; style-src 'self' 'unsafe-inline' https://cdn.jsdelivr.net; connect-src 'self' * https://*.$kymadomain wss://*.$kymadomain; font-src 'self' data:; frame-ancestors https://*.$kymadomain; object-src 'none'; media-src 'self'; form-action 'self'; img-src * data:; child-src * blob:; worker-src 'self' blob: data:;";
```
parostatkiem, line 46: I think you can remove the other two domains since the `*` covers it all.

kwiatekus: true, but I wanted to keep them as we will go back to those more restrictive connect-src policy as soon all MPS UI will be available from one central place. We wont need it anymore and could keep to the previous setup (incl. connecting to 'self' and resources from within $kymadomain)

PR #1729, opened 2020-04-22
Support allow access strategy in API Rules UI

kwiatekus:
**Description**

- Support `allow` strategy type for API Rules UI

### DIFF
--- a/core-ui/src/components/ApiRules/AccessStrategy/AccessStrategy.js
+++ b/core-ui/src/components/ApiRules/AccessStrategy/AccessStrategy.js
@@ -1,24 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Badge, Icon } from 'fundamental-react';
-
-const passAll = {
-  value: 'allow',
-  displayName: 'Allow',
-};
-const jwt = {
-  value: 'jwt',
-  displayName: 'JWT',
-};
-const noop = {
-  value: 'noop',
-  displayName: 'noop',
-};
-const oauth2 = {
-  value: 'oauth2_introspection',
-  displayName: 'OAuth2',
-};
-const accessStrategiesList = [passAll, noop, oauth2, jwt];
+import classNames from 'classnames';
+import accessStrategyTypes, { usesMethods } from '../accessStrategyTypes';
 
 const AccessStrategy = ({ strategy }) => {
   const selectedType = strategy.accessStrategies[0].name;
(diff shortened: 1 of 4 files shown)


core-ui/src/components/ApiRules/ApiRuleForm/AccessStrategyForm/AccessStrategyForm.js
```diff
@@ -20,7 +20,10 @@ const AVAILABLE_METHODS = ['GET', 'POST', 'PUT', 'DELETE'];
 // const URLregexp = new RegExp(
```
sjanota, line 20: Could you remove that commented regexp?

core-ui/src/components/ApiRules/ApiRuleForm/AccessStrategyForm/AccessStrategyForm.js
```diff
@@ -195,3 +170,39 @@ function OAuth2Details({ config, setConfig }) {
     />
   );
 }
+
+function HTTPMethods({ strategy, setStrategy }) {
```
sjanota, line 174: Not a blocker, but I'd change this component so it is unaware of strategy.

core-ui/src/components/ApiRules/accessStrategyTypes.js
```diff
@@ -0,0 +1,23 @@
...
+const jwt = {
+  value: 'jwt',
+  displayName: 'JWT',
+};
+const oauth2_introspection = {
+  value: 'oauth2_introspection',
+  displayName: 'OAuth2',
+};
+
+const accessStrategyTypes = { allow, noop, oauth2_introspection, jwt };
```
dariadomagala, line 19: I think that we should set `noop` as the first entry since its a default value.

core-ui/src/components/ApiRules/AccessStrategy/AccessStrategy.js
```diff
@@ -30,25 +14,27 @@ const AccessStrategy = ({ strategy }) => {
...
-              accessStrategiesList.find(item => item.value === selectedType)
-                .displayName
-            }
+            {accessStrategyTypes[selectedType].displayName}
           </Badge>
         </div>
-        <div className="methods">
+        <div
+          className={classNames('methods', {
+            'fd-hidden': !usesMethods(selectedType),
```
dariadomagala, line 30: why do we set it as 'hidden' instead of just not adding this div?

kwiatekus: as it makes the UI jump. with the invisible elements we managed to keep the positions as the boxes are kept

PR #1768, opened 2020-05-05
Use "services" as default tab in Catalog and Instances

Wawrzyn321:
**Description**

- Use "services" as default tab

**Related issue(s)**
Fixes [345](https://github.tools.sap/kyma/backlog/issues/345)


kwiatekus:
Would be more natural if `Services` tab show up as a first tab as it will be selected by default
Screen Shot 2020-05-06 at 09 09 43

PR #1787, opened 2020-05-12
Patch core

sjanota:
**Description**

- Adapt core Docekrfile nad Makefile so single `release` target is exposed for CI

**Related issue(s)**


sjanota:
I moved all of the building into Dockerfile.

PR #1796, opened 2020-05-15
Increase default max memory per container

kwiatekus:
**Description**

Default memory limit (1Gi) is not enough for serverless to build function images.
[Job building images requires 1100Mi](https://kyma-project.io/docs/components/serverless/#configuration-serverless-chart). 
The change is not enough. 
As a follow up @michal-hudy proposed  informing user about insufficient resources in the  serverless ui

- increase default value to 1100Mi


core/src/app/content/namespaces/namespace-create/namespace-create.component.ts
```diff
@@ -172,7 +172,7 @@ export class NamespaceCreateComponent {
     this.labels = ['istio-injection=enabled'];
     this.memoryLimits = '3Gi';
     this.memoryRequests = '2.8Gi';
-    this.max = '1Gi';
+    this.max = '1100Mi';
```
parostatkiem, line 175: I don't think this file is used anywhere, is it? Please check it and remove if needed.

PR #1816, opened 2020-05-26
Unify statuses on lists

Wawrzyn321:
**Description**

- Modify StatusBadge component to match the design and allow to use _Tooltip_
- use `html` prop instead of `title` in Popper to allow passing any React node
- rename `title` prop to `content` in the _Tooltip_ component to be more descriptive
- use the StatusBadge on all React lists

**Related issue(s)**


shared/components/StatusBadge/StatusBadge.js
```diff
@@ -1,24 +1,73 @@
...
+
+export const StatusBadge = ({
+  status,
+  type,
+  children,
+  autoResolveType = false,
+  tooltipProps = {},
+  className,
+}) => {
+  if (autoResolveType) type = resolveType(status);
```
sjanota, line 41: Another approach might be to auto-resolve whenever `type` is empty. Have you considered it?

shared/components/StatusBadge/StatusBadge.js
```diff
@@ -1,24 +1,73 @@
...
+    className,
+  );
+
+  const badgeElement = (
+    <Badge role="status" modifier="filled" className={classes}>
+      {status}
+    </Badge>
+  );
+
+  return children ? (
```
sjanota, line 58: Purity comment: we use `children` here as an optional property to provide a tooltip. However, I'd expect `children` to hold the main content of the component (like status), not side content. The API is nice, but may not be intuitive.

shared/components/StatusBadge/test/StatusBadge.test.js
```diff
@@ -1,11 +1,13 @@
...
 import React from 'react';
 import { StatusBadge } from '../StatusBadge';
+import { render } from '@testing-library/react';
 
 describe('StatusBadge', () => {
-  it('should render loading state initially', () => {
-    const component = renderer.create(<StatusBadge status="INITIAL" />);
-    const tree = component.toJSON();
-    expect(tree).toMatchSnapshot();
+  it('renders status text with proper role', () => {
```
sjanota, line 6: Tooltip test?

sjanota, line 6: Auto-resolve is not tested at all. We should have at least simple test to cover it too.

Wawrzyn321: We tried, but even with forced trigger (`trigger={'manual'}, open={true}`) Tooltip content won't be visible in jsdom.
That's why we've removed Maciek's tests, as they only tested existence of Tooltip's wrapper over tooltipable element, but not the tooltip popup itself.

shared/components/StatusBadge/StatusBadge.js
parostatkiem: `undefined` is one of five possible types, the grey one. I get your point but see no benefit over the current approach. Both are correct in my opinion.

shared/components/StatusBadge/test/StatusBadge.test.js
parostatkiem: That's another problem with the testing framework - we **can't**  actually get the component `background-color` property although both _testing-library_ and the _jsdom_ (`getComputedStyles()`) theoretically should make it possible.

sjanota: Anyway, I'd leave here at least some trace so if something changes accidentally (or fundamental messes the API again) tests will tell us about it.

sjanota: Can we maybe render two elements, one with autoResolve and one without and assert that they have different styles? That should be sufficient I think.

shared/components/StatusBadge/StatusBadge.js
```diff
@@ -6,6 +6,11 @@ import classNames from 'classnames';
 import { Tooltip } from '../Tooltip/Tooltip';
 
 const resolveType = status => {
+  if (typeof status !== 'string') {
+    console.warn(`'autoResolveType' prop requires a non-string status value`);
```
sjanota, line 10: It's inaccurate. It requires a string value.

sjanota: And it's already handled by Props, right? Why another check?

shared/components/StatusBadge/test/StatusBadge.test.js
Wawrzyn321: Tooltip used here is from `react-tippy`, not from Fundamental, so it's less prone to changes. If `Badge` API changes, then `renders status text with proper role` should do.

shared/components/StatusBadge/StatusBadge.js
sjanota: You should be able to set string PropType on children

PR #1818, opened 2020-05-27
Unify angular list statuses

parostatkiem:
**Description**

- modify status-label component to work better and look like it should
- use it on all Angular lists
- ...

**Related issue(s)**

### DIFF
--- a/core/src/app/content/namespaces/operation/services/service-details/service-api-rules/service-api-rules.component.html
+++ b/core/src/app/content/namespaces/operation/services/service-details/service-api-rules/service-api-rules.component.html
@@ -11,7 +11,10 @@
   <tbody fd-table-body *ngIf="!error">
     <tr *ngFor="let apiRule of apiRules" fd-table-row>
       <td class="fd-has-font-weight-bold">
-        <a href="javascript:void(null)" (click)="navigateToDetails(apiRule.name)">
+        <a
+          href="javascript:void(null)"
+          (click)="navigateToDetails(apiRule.name)"
+        >
           {{ apiRule.name }}
         </a>
       </td>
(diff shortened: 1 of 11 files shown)


core/src/app/shared/components/status-label/status-label.component.scss
```diff
@@ -1,15 +1,65 @@
-.status-label {
-  text-align: left;
-  text-transform: uppercase;
-  position: relative;
-  cursor: default;
-}
+// .status-label {
```
dariadomagala, line 1: why not just removing these classes?

core/src/app/shared/components/status-label/status-label.component.ts
```diff
@@ -1,29 +1,37 @@
-import { Component, OnInit, Input } from '@angular/core';
-import { TooltipComponent } from '../tooltip/tooltip.component';
+import { Component, Input } from '@angular/core';
+
+type Status = 'warning' | 'error' | 'info' | 'success' | 'ok' | undefined;
```
dariadomagala, line 3: are you sure those are the only possibilities?

core/src/app/content/namespaces/operation/services/service-details/service-api-rules/service-api-rules.component.html
```diff
@@ -11,7 +11,10 @@
   <tbody fd-table-body *ngIf="!error">
     <tr *ngFor="let apiRule of apiRules" fd-table-row>
       <td class="fd-has-font-weight-bold">
-        <a href="javascript:void(null)" (click)="navigateToDetails(apiRule.name)">
+        <a
```
dariadomagala, line 14: do we even need API Rules anymore? we use core-ui views

parostatkiem: Yes, we do. It's a list of API Rules assigned to this service. You can create these with the new core-ui view.

core/src/app/shared/components/status-label/status-label.component.ts
```diff
@@ -1,29 +1,37 @@
...
+    // Angular explodes once the given classes string can't be split by spaces nicely. Thank you Angular.
+    return `status-badge ${this.getStatusClass(this.statusType)} ${
+      this.description ? 'has-tooltip' : ''
+    }`.trim();
+  }
 
-  public getStatusClass = (type: string) => {
+  public getStatusClass = (type: Status) => {
     switch (type) {
       case 'warning':
```
dariadomagala, line 22: We should also add 'terminated/terminating here' because I don't think it's supposed to be green.
Screen Shot 2020-05-28 at 09 00 02 

And it would be good make the default option grey.

parostatkiem: But the default option **is** grey.

Issue #1824, opened 2020-06-02
Kyma Log UI log search field doesn't allow all LogQL queries 


hisarbalik:
**Description**
Console log ui search field validation doesn't allow all LogQL queries to enter, queries like ```{stream=~"stderr|stdout"}```

**Expected result**
Log ui in the console should allow all LogQL queries and able to send loki log server

**Actual result**

In the log ui "advance settings"  currently entering queries like ``` {stream=~"stderr|stdout"}``` not possible due to form validation.

**Steps to reproduce**

- Go to  kyma console log ui

- Click "Advance Settings" and enter query ```{stream=~"stderr|stdout"}```

kwiatekus:
Fixed with https://github.com/kyma-project/console/pull/1825

PR #1825, opened 2020-06-02
Fix validation error for logQL regex in log ui advanced search

kwiatekus:
**Description**

- Support regex in log ui advanced mode 

**Related issue(s)**
https://github.com/kyma-project/console/issues/1824

### DIFF
--- a/logging/src/components/Shared/QueryInput/QueryInput.js
+++ b/logging/src/components/Shared/QueryInput/QueryInput.js
@@ -2,7 +2,7 @@ import React, { useState } from 'react';
 import { FormInput } from 'fundamental-react';
 
 const labelRegexp = new RegExp(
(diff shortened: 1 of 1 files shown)


logging/src/components/Shared/QueryInput/QueryInput.js
```diff
@@ -1,7 +1,9 @@
 import React, { useState } from 'react';
 import { FormInput } from 'fundamental-react';
 
-const labelRegexp = new RegExp(/[a-z0-9A-Z-_.]+="[a-z0-9A-Z-_.]+"/);
+const labelRegexp = new RegExp(
+  /[a-z0-9A-Z-_.]+(!?=|=~|!~)"[a-z0-9A-Z-_.|*+\\\[\]]+"/,
```
Wawrzyn321, line 5: Expression won't work for query that uses ``` ` ``` instead of `"`, for example ```{name!~`mysql-\d+`}``` from [Loki's README](https://github.com/grafana/loki/blob/master/docs/logql.md#user-content-log-stream-selector).

PR #1836, opened 2020-06-08
Add proper handling for k8s label schema

### LINKED ISSUE #1766: Cannot label namespace as supported by kubernetes
**Description**
I have several namespaces labeled with `kiali.io/member-of=kyma-system`, however in the kyma console I cannot label a namespace with that label

Screenshot 2020-04-30 at 22 10 09

**Expected Result**
I have full label syntax available as with kubernetes itself.

**Actual Result**
Syntax is restricted for unknown reasons

Wawrzyn321:
**Description**

- Add handling for key prefix

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/

**Related issue(s)**
Fixes #1766

### DIFF
--- a/core/src/app/shared/components/labels-input/labels-input.component.spec.ts
+++ b/core/src/app/shared/components/labels-input/labels-input.component.spec.ts
@@ -186,7 +186,7 @@ describe('LabelsInputComponent', () => {
       it('if valid format and invalid character in key, sets message and return true', () => {
         const label = 'öö=val1';
         const result: boolean = component['setWrongLabelMessage'](label);
-        const expected = `Invalid key öö! Key consists of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character. It can be prefixed with DNS domain, separated with "/".`;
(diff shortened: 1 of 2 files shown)


core/src/app/shared/components/labels-input/labels-input.component.spec.ts
```diff
@@ -158,12 +158,63 @@ describe('LabelsInputComponent', () => {
...
-    it('if valid format and invalid characters, sets message and return true', () => {
-      const label = 'öö=val1';
-      const result: boolean = component['setWrongLabelMessage'](label);
-      const expected = `Invalid label öö=val1! In a valid label, a key cannot be empty, a key/value consists of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character.`;
-      expect(component.wrongLabelMessage).toBe(expected);
-      expect(result).toBe(true);
+    describe('key', () => {
+      it('if multiple "/", sets message and return true', () => {
+        const label = 'a/b/c=d';
+        const result: boolean = component['setWrongLabelMessage'](label);
```
parostatkiem, line 164: Is there any reason why you use a dynamic property name?

Wawrzyn321: That method is private. Of course, we could rerender the component for every test case, but a) behaviour is already covered above b) it's just faster.

PR #1840, opened 2020-06-12
Update serverless domain to use envs from configmap

aerfio:
**Description**

Changes proposed in this pull request: unhardcode restriced envs and read them from configmap

- Update serverless domain to use envs from configmap
- we are not reading min resources from configmap _yet_, as it would mean we would need to change a lot of files (basically we read them in normal js functions, not inside any component, so we can't use `useQuery` hook, and we do not export apollo-client instance :sad:)

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/8780


aerfio:
Already taken care of in other PR

PR #1844, opened 2020-06-15
Update Luigi everywhere

parostatkiem:
**Description**

Each update has been set to the newest version `1.2.1`

- Update Core and Client in `/core`
- update Client globally
- Update Core and Client in `/compass`

**Related issue(s)**


tests/helpers.js
```diff
@@ -26,7 +26,7 @@ const retry = async (t, retries, func, message) => {
 export const findActiveFrame = async t => {
   return await retry(
     t,
-    3,
+    5,
```
dariadomagala, line 29: Please change it back to 3

PR #1880, opened 2020-07-06
Fix create ns modal test

parostatkiem:
**Description**

- rewrite the problematic test using @testing-library/react
- ...
- ...

**Related issue(s)**


core-ui/src/components/CreateNamespaceForm/test/CreateNamespaceForm.test.js
```diff
@@ -1,6 +1,12 @@
-import React from 'react';
+import React, { memo } from 'react';
```
Wawrzyn321, line 1: `Memo` isn't used anywhere. Same goes for `waitForDomChange`.

PR #1888, opened 2020-07-14
Use new success notifications in Addons

parostatkiem:
**Description**

- ...
- ...
- ...

**Related issue(s)**


add-ons/src/services/Subscriptions.service.ts
```diff
@@ -55,12 +57,12 @@ interface AddonsConfigurationSubscriptionVariables {
 const useSubscriptions = () => {
   const { namespaceId: currentNamespace } = useContext(LuigiContext);
   const { setOriginalConfigs } = useContext(ConfigurationsService);
-  const { successNotification, errorNotification } = useContext(
-    NotificationsService,
-  );
+  const { errorNotification } = useContext(NotificationsService); // old notifications
```
dariadomagala, line 60: Can't we use new notifications also for errors?

parostatkiem: Well, I think we can

PR #1901, opened 2020-07-22
Fix small text issues, wrap notification text

Wawrzyn321:
**Description**

It turns out "Remove Applications and IDPPresets from core-ui" (165d891f) was the culprit - it removed `ApiDetails`, which in turn included `GenericComponent`, along with its styles. Without them, some texts' fonts are broken.

If you insert ```
import { GenericComponent } from '@kyma-project/generic-documentation';``` (on master) expected text sizes are back.

- Fix sizing of text across the Console
- Wrap text in success notification (also make it a bit wider)
- Remove `FileInput` from Compass and use `shared` one instead, as they only differed in named export.
- Remove redundant console.log from compass
- Add timeout of 10s for catalog test, as it started failing on Prow, while locally it was passing.

**Related issue(s)**


shared/components/StatusBadge/StatusBadge.scss
```diff
@@ -1,5 +1,5 @@
 .fd-badge.status-badge {
-  font-size: var(--sapFontSize, 0.875rem);
+  font-size: var(--sapFontSize, 12.25px);
```
dariadomagala, line 2: Where is this value from?

Wawrzyn321: Default font size is 14px, 14px * 0.875 = 12.25px. In some instances using _rem_ yielded tiny status texts (APIRules list), in others usage of _em_ resulted in different size of status texts (Cluster Addons), so I decided to go for absolute units.

Issue #1853, opened 2020-06-18
Failing smoke tests


Wawrzyn321:
**Description**
See comments.

Wawrzyn321:
08.06
Catalog view is rendered

   1) The specified selector does not match any element in the DOM tree.

Wawrzyn321:
10.06
 Docs view is rendered
  
     1) The specified selector does not match any element in the DOM tree.
  
         > | Selector('iframe')
10.06 post-master-kyma-gke-integration
Deployments view is rendered

   1) AssertionError: expected false to be truthy

      Browser: Firefox 68.0 / Linux 0.0

         53 |  //THEN
         54 |  await findActiveFrame(t);
10.06 post-master-kyma-gke-upgrade
Switching the iframe failed. Retries left: 2
Switching the iframe failed. Retries left: 1
Switching the iframe failed. Retries left: 0
 ✖ Catalog view is rendered

   1) The specified selector does not match any element in the DOM tree.

       > | Selector('iframe')
         |   .filterVisible()

      Browser: Firefox 68.0 / Linux 0.0
10.06 TO JEST ŚWIEŻE https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/post-master-kyma-gke-integration/1270669897982021632
10.06 z PRa
Switching the iframe failed. Retries left: 2
  Switching the iframe failed. Retries left: 1
  Switching the iframe failed. Retries left: 0
   ✖ Catalog view is rendered
  
     1) The specified selector does not match any element in the DOM tree.
  
         > | Selector('iframe')
           |   .filterVisible()
  
        Browser: Firefox 68.0 / Linux 0.0
10.06
Switching the iframe failed. Retries left: 2
Switching the iframe failed. Retries left: 1
Switching the iframe failed. Retries left: 0
 ✖ Docs view is rendered

   1) The specified selector does not match any element in the DOM tree.

       > | Selector('iframe')
         |   .filterVisible()

      Browser: Firefox 68.0 / Linux 0.0
1

Wawrzyn321:
12.06
  Switching the iframe failed. Retries left: 2
  Switching the iframe failed. Retries left: 1
  Switching the iframe failed. Retries left: 0
   ✖ Catalog view is rendered
  
     1) The specified selector does not match any element in the DOM tree.
  
         > | Selector('iframe')
           |   .filterVisible()
  
        Browser: Firefox 68.0 / Linux 0.0
[...]

PR #1903, opened 2020-07-24
Fix smoke tests (brave statement)

parostatkiem:
**Description**

- adjust timeouts to smaller values
- decrease tests speed to 0.1
- disable Docs test
- enable Deployments test
- wait for a second before another retry
- make dockerfile more independent from external projects 
- make docker image a bit smaller
- increase retries to 5
- update _testcafe_ to the newest version

**Related issue(s)**
#1853


tests/helpers.js
```diff
@@ -10,7 +22,7 @@ export const testIf = (condition, testName, testToRun) => {
   }
 };
 
-const retry = async (t, retries, func, message) => {
+const retry = async (t, retries, func, message, waitAfterFail = 3000) => {
   try {
```
Ogurecher, line 26:
```suggestion
  try {
    reloadPage();
```

tests/helpers.js
```diff
@@ -19,15 +31,18 @@ const retry = async (t, retries, func, message) => {
...
     );
     if (retries === 1) throw err;
-    return await retry(t, retries - 1, func, message);
+    await new Promise(res => setTimeout(res, waitAfterFail));
+    return await retry(t, retries - 1, func, message, waitAfterFail);
   }
 };
 
 export const findActiveFrame = async t => {
+  reloadPage();
```
Ogurecher, line 40:
```suggestion

```

parostatkiem:
@Ogurecher Thanks for the advice.
One question - why do you thing this will change anything? 
I tried adding it at the beginning of the `retry` as well as at the end (before the 3s artificial wait). Absolutely nothing changed :(

Ogurecher:
@parostatkiem,

Switching to the iframe fails inconsistently, and this makes me think that failures occur for one of these reasons:

* The iframe failed to load or was loaded incorrectly (page reload could fix it)
* The iframe hasn't loaded yet (this may be fixed if you wait for it to load)


I suggested reloading the page before each retry to address the first reason - multiple reloads should have fixed the issue.
However, as I see that this is not the case, I think that we should try to cover both these cases:

* Increase `waitAfterFail` and `retries`. For example, set `waitAfterFail = 10000` and `retries = 12`
* Reload the page after every n'th retry fails (in the `catch` block, before `setTimeout`). To start, you can set `n = 3`.


It should fix the issue or at least reduce the failure rate. Should you need help, I can add these changes as suggestions to your PR.

Please let me know if this resolves your issue.

parostatkiem:
@Ogurecher the test failed on all 4 jobs 🙂 
Any other ideas?

tests/helpers.js
```diff
@@ -10,7 +22,7 @@ export const testIf = (condition, testName, testToRun) => {
   }
 };
 
-const retry = async (t, retries, func, message) => {
+const retry = async (t, retries, func, message, waitAfterFail = 10000) => {
```
Ogurecher, line 25: I'd suggest using loops instead of recursive calls as it improves code readability. It's just a codestyle nitpick though, feel free to leave this part as it is.
If you don't want to change this part, make sure that you await the `reloadPage()` call.

```js
const retry = async (t, retries, func, message, waitAfterFail = 10000) => {
  for (let i = 0; i < retries; i++) {
    try {
      return await func(t);
    } catch (err) {
      console.log(
        `${message ? message : 'Testing iframe'} failed. Retries left: ${retries -
          1}`,
      );
      if (i === retries - 1) throw err;
      if (i % 3 === 0) await reloadPage();
      await new Promise(res => setTimeout(res, waitAfterFail));
    }
  }
};
```

parostatkiem:
@Ogurecher It not only didn't help but also a weird error appeared
image

tests/helpers.js
```diff
@@ -2,6 +2,18 @@ import config from './config';
 import { ClientFunction, Selector, Role } from 'testcafe';
 import chalk from 'chalk';
 
+const getLocation = ClientFunction(() => window.location.pathname);
+export const reloadPage = ClientFunction(() => window.location.reload(true));
```
Ogurecher, line 6:
```suggestion
export const reloadPage = ClientFunction(() => { window.location.reload(true) });
```

PR #1934, opened 2020-08-17
Window titles

Wawrzyn321:
**Description**

- Add a hook for setting window title, and use it according to [6430](https://github.com/kyma-project/kyma/issues/6430).
- As of current Luigi, setting the window title also sets the top-left header text. As a workaround, we decided to just hide it with CSS.
- I also tried to use the hook (from `shared`) in `content-ui`, but importing `shared` failed with
```
./src/react-shared/index.js
  Line 4:42:  'define' is not defined  no-undef
  Line 5:3:   'define' is not defined  no-undef
```
, even though import worked fine
```
import * as ReactShared from '../react-shared';
console.log(!!ReactShared.withTitle)
```
printing out _true_ to dev console. As I only needed one tiny function, I decided to just copy the body into one content UI component.
- I didn't change the Compass UI titles, as they already have "Management Plane" title.

**Related issue(s)**


akucharska (commented):
Very nice job. 
Below one suggestion

akucharska (approved):
LGTM

majakurcius (approved):

magicmatatjahu (approved):

PR #1948, opened 2020-08-26
Addons ui fixes

akucharska:
**Description**

- API rule list & detail: remove border from link
- API rule list: action should always be in one line
- fix colors of checked checkboxes

**Related issue(s)**


Wawrzyn321 (changes requested):
Nice! I think we could promote the `.no-border` class to the `index.scss` file, where `.link` itself is defined.

Wawrzyn321 (approved):

Issue #738, opened 2019-04-02
Use GraphQL queries for rendering secrets in Console UI


kwiatekus:
**Description**
Manage secrets in console UI via console-backend

**AC**
 - List secrets in the given namespace using graphQL queries ( subscriptions not required).
 - Use them as well to render secret details view.
 - Use mutations to delete or update the secret value

Please consider also:
 - rewriting UI to react JS (list view and secret details)
 - getting rid of filtering by labels if it complicates the implementation
 - including the task about decoding secret data : https://github.com/kyma-project/console/issues/727
 - putting metadata of secret into page header

**Reasons**
We should use CBS API everywhere in the UI. Currently we have a mix - we should use only one proxy.

kwiatekus:
Duplicated by https://github.com/kyma-project/console/issues/932

PR #1967, opened 2020-09-09
Rewrite secrets

akucharska:
**Description**

- Rewrite secrets list to react
- Rewrite secrets details to react

**Related issue(s)**

https://github.com/kyma-project/console/issues/738


Wawrzyn321:
I think annotations in `key=value` (as seen in labels) format would be nicer than a key and a modal with value.

Wawrzyn321:
What's more, we should add a validation for editor values, such as non-parseable JSON or non-object values, and then control the "Save" button disabledness accordingly.

Wawrzyn321:
By the way I'm wondering if displaying the annotations is required at all. I know the original task mentioned _putting metadata of secret into page header_, but on other views we are just displaying labels. Do we know that annotations are more valuable than labels?

dariadomagala (approved):

PR #1971, opened 2020-09-10
Fetch services via cbs

dariadomagala:
**Description**

- Fetch services via cbs

**Related issue(s)**


core/src/luigi-config/navigation/static-navigation-model.js
```diff
@@ -134,16 +134,16 @@ export function getStaticChildrenNodesForNamespace(){
       pathSegment: 'services',
       navigationContext: 'services',
       label: 'Services',
-      viewUrl: '/consoleapp.html#/home/namespaces/:namespaceId/services',
+      viewUrl: config.coreModuleUrl + '/home/namespaces/:namespaceId/services',
       keepSelectedForChildren: true,
+      viewGroup: '_core_ui_',
```
Wawrzyn321, line 139:
```suggestion
      viewGroup: coreUIViewGroupName,
```

core-ui/src/components/Services/ServicesList/ServicesList.js
```diff
@@ -0,0 +1,72 @@
...
+    listOfEndpoints(entry),
+    <Moment unix fromNow>
+      {entry.creationTimestamp}
+    </Moment>,
+    <Labels labels={entry.labels} />,
+  ];
+
+  return (
+    <GenericList
+      title="Services"
```
Wawrzyn321, line 62: I'd say this title isn't needed there. We have already a title with resource type in the header (like in API Rules or OAuth Clients).

core-ui/src/components/Services/ServicesList/ServicesList.js
```diff
@@ -0,0 +1,72 @@
...
+      name: 'Details',
+      handler: navigateToServiceDetails,
+    },
+  ];
+
+  const listOfEndpoints = entry => {
+    if (entry.ports && entry.ports.length) {
+      return entry.ports.map(port => {
+        const portValue = `${entry.name}.${namespace}: ${port.port} ${port.serviceProtocol}`;
+        return <div key={portValue}>{portValue}</div>;
```
Wawrzyn321, line 35: I don't know if we care about the semantics this much, but it would be nice to use `ul` and `li` here.

core-ui/src/components/Services/ServicesList/ServicesList.js
```diff
@@ -0,0 +1,72 @@
...
+import { GenericList, Labels, EMPTY_TEXT_PLACEHOLDER } from 'react-shared';
+
+import { GET_SERVICES } from 'gql/queries';
+
+ServicesList.propTypes = { namespace: PropTypes.string.isRequired };
+
+export default function ServicesList({ namespace }) {
+  const navigateToServiceDetails = service =>
+    LuigiClient.linkManager().navigate(
+      `details/${encodeURIComponent(service.name)}`,
```
Wawrzyn321, line 16: Encoding isn't needed here. First, it isn't even handled on details view. And second, k8s won't allow an unsafe service name.

Wawrzyn321 (changes requested):
See comments.

Wawrzyn321 (approved):

Issue #1897, opened 2020-07-20
POC : YAML preview for API Detail View


kwiatekus:
**Description**

Allow to display API Rule details in form of yaml content.
This POC should evaluate possible options, ie:
 - (blocking/resizable) drawer 
 
In this POC yaml edit-mode is not required. Read-only mode.
Use monaco editor to achieve the same feature set as in function UI (collapsibility, syntax colouring).

**Reasons**

Promote educational aspects of console UI. [CNCF developers](https://github.com/kyma-project/community/blob/master/manifesto/personas.md#matthew---cloud-native-software-engineer) would like to learn kyma API.
Showing YAML content of displayed items helps understanding kyma API

PR #2001, opened 2020-09-23
Implement side drawer with yaml

akucharska:
**Description**

- Implement side drawer with yaml

**Related issue(s)**

https://github.com/kyma-project/console/issues/1897

### DIFF
--- a/core-ui/src/components/ApiRules/ApiRuleDetails/ApiRuleDetails.js
+++ b/core-ui/src/components/ApiRules/ApiRuleDetails/ApiRuleDetails.js
@@ -82,7 +82,7 @@ function EditButton({ apiRuleName }) {
   return (
     <Button
       onClick={() => navigateToEditView(apiRuleName)}
-      option="light"
+      option="emphasized"
       aria-label="edit-api-rule"
     >
       Edit
@@ -109,6 +109,7 @@ const YamlContent = yaml => {
         language={'json'}
         theme="vs-light"
         value={yaml}
+        options={{ readOnly: true }}
       />
     </>
   );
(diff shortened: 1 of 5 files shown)


shared/components/SideDrawer/SideDrawer.js
```diff
@@ -1,19 +1,23 @@
...
+  hideDefaultButton,
 }) => {
   return (
     <div className={classNames('side-drawer', { 'side-drawer--open': isOpen })}>
       {(isOpen || children) && (
-        <button className="open-btn" onClick={() => setOpen(!isOpen)}>
+        <button
+          className={`open-btn ${hideDefaultButton ? 'open-btn-hidden' : ''}`}
+          onClick={() => setOpen(!isOpen)}
+        >
```
Wawrzyn321, line 20: Why don't we use just `{!hideDefaultButton && <button className="open-btn" onClick=...` instead of setting CSS visibility?

core-ui/src/components/ApiRules/ApiRuleDetails/ApiRuleDetails.js
```diff
@@ -92,7 +98,23 @@ function navigateToEditView(apiRuleName) {
 
 const breadcrumbItems = [{ name: 'API Rules', path: '/' }, { name: '' }];
 
-function ApiRuleDetailsHeader({ apiRule }) {
+const YamlContent = yaml => {
+  yaml = JSON.stringify(yaml, null, 1);
+  return (
+    <>
+      <h1 className="fd-has-type-4">YAML</h1>
+      <ControlledEditor
```
Wawrzyn321, line 106: I think in this case the editor could be readonly.

Wawrzyn321 (changes requested):
See comments. What's more, I think the "YAML" button should use "light" variant, just like other header actions buttons.
Would you kindly move "Delete" button to the right and make "Edit" button emphasised, to keep it consistent with other views?

shared/components/SideDrawer/SideDrawer.js
akucharska: Because without button part of the drawer is visible at all times

Wawrzyn321 (approved):

PR #2010, opened 2020-10-02
Remove smoke tests for logs

akucharska:
**Description**

- Remove LOG UI from Kyma installation in favour of deep-linking to grafana UI

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/9196

### DIFF
--- a/core-ui/src/components/NamespaceDetails/NamespaceDetailsHeader/NamespaceDetailsHeader.js
+++ b/core-ui/src/components/NamespaceDetails/NamespaceDetailsHeader/NamespaceDetailsHeader.js
@@ -16,10 +16,7 @@ export default function NamespaceDetailsHeader({ namespace }) {
   const [deleteNamespace] = useMutation(DELETE_NAMESPACE);
 
   const DOMAIN = useConfig().fromConfig('domain');
-
-  /* eslint-disable no-useless-escape */
-  const query = `{namespace=\"${namespace.name}\"}`;
-  /* eslint-enable no-useless-escape */
+  const query = `{namespace="${namespace.name}"}`;
 
   const handleNamespaceDelete = () =>
     handleDelete(
--- a/core-ui/src/components/NamespaceDetails/NamespaceDetailsHeader/NamespaceDetailsHeader.js
+++ b/core-ui/src/components/NamespaceDetails/NamespaceDetailsHeader/NamespaceDetailsHeader.js
@@ -18,7 +18,9 @@ const DOMAIN = getApiUrl('domain');
 export default function NamespaceDetailsHeader({ namespace }) {
   const [deleteNamespace] = useMutation(DELETE_NAMESPACE);
 
+  /* eslint-disable no-useless-escape */
   const query = `{namespace=\"${namespace.name}\"}`;
+  /* eslint-enable no-useless-escape */
 
   const handleNamespaceDelete = () =>
     handleDelete(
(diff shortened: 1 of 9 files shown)


shared/components/LogsLink/LogsLink.js
```diff
@@ -0,0 +1,33 @@
...
+  domain,
+  query,
+  from = 'now-1h',
+  to = 'now',
+  dataSource = 'Loki',
+  mode = 'Logs',
+}) => {
+  if (!domain) {
+    return null;
+  }
```
Wawrzyn321, line 14: I think we should add a `required` propType for the _domain_ (and maybe other props), as just rendering nothing if domain is falsy may be confusing.

core-ui/src/components/NamespaceDetails/NamespaceDetailsHeader/NamespaceDetailsHeader.js
```diff
@@ -4,17 +4,24 @@ import PropTypes from 'prop-types';
...
 
+const DOMAIN = getApiUrl('domain');
+
 export default function NamespaceDetailsHeader({ namespace }) {
   const [deleteNamespace] = useMutation(DELETE_NAMESPACE);
 
+  /* eslint-disable no-useless-escape */
+  const query = `{namespace=\"${namespace.name}\"}`;
+  /* eslint-enable no-useless-escape */
+
```
Wawrzyn321, line 24: Why is the escape required? I removed it, and the link worked equally well.

core-ui/src/components/NamespaceDetails/NamespaceDetailsHeader/NamespaceDetailsHeader.js
```diff
@@ -4,17 +4,24 @@ import PropTypes from 'prop-types';
 
 import { useMutation } from '@apollo/react-hooks';
 import { DELETE_NAMESPACE } from 'gql/mutations';
+import { getApiUrl } from '@kyma-project/common';
```
Wawrzyn321, line 7: It'd be better to use `useConfig().fromConfig('domain')` instead of `getApiUrl`.

Wawrzyn321 (approved):
I've left some comments for the code, but anyway, it works really good!

Wawrzyn321 (changes requested):
...

core-ui/src/components/NamespaceDetails/NamespaceDetailsHeader/NamespaceDetailsHeader.js
akucharska: Without it job fails because of unnecessary \.

Wawrzyn321: Yup, but don't we use just ``{namespace="${namespace.name}"}``?

shared/components/LogsLink/LogsLink.js
```diff
@@ -0,0 +1,40 @@
...
+  const grafanaLink = `https://grafana.${domain}/explore?left=${JSON.stringify(
+    parameters,
+  )}`;
+
+  return <Link className="fd-button" url={grafanaLink} text="Logs" />;
+};
+
+LogsLink.propTypes = {
+  domain: PropTypes.string.isRequired,
+  query: PropTypes.string.isRequired,
```
Wawrzyn321, line 35: Query shouldn't be `required`, if we are testing it in line 13.

Wawrzyn321 (changes requested):

Wawrzyn321 (approved):

PR #2034, opened 2020-10-15
Resources on namespace view

parostatkiem:
**Description**

- Add `YamlEditorContext` context
- Add `LimitRanges` and `ResourceQuotas` components
- Remove the whole `Resources` view including all the code behind it
- Display resources in the Namespace Details view (using fancy grid)

**Related issue(s)**


core-ui/src/gql/queries.js
```diff
@@ -1,5 +1,9 @@
 import gql from 'graphql-tag';
 
+// const LIMITS_FRAGMENT = `
+// resources
+// `
+
```
Wawrzyn321, line 6: I believe it isn't needed. Or...
```
const LIMITS_FRAGMENT = gql`
  fragment limitsFragment on ResourceLimits {
    cpu
    memory
  }
`
```

core-ui/src/components/NamespaceDetails/ResourceQuotas/ResourceQuotas.js
```diff
@@ -0,0 +1,84 @@
...
+
+const ResourceQuotas = ({ resourceQuotas, namespaceName: namespace }) => {
+  const setEditedJson = useYamlEditor();
+  const notificationManager = useNotification();
+  const editedResourceQuota = useRef(null);
+
+  const [updateResourceQuota] = useMutation(UPDATE_RESOURCE_QUOTA, {
+    onCompleted: ({ updateResourceQuota }) =>
+      notificationManager.notifySuccess({
+        content: formatMessage('Succesfully updated', updateResourceQuota.name),
```
Wawrzyn321, line 35: The message displayed is "Succesfully updated", without the quota name.

core-ui/src/components/NamespaceDetails/NamespaceDetails.js
```diff
@@ -33,10 +37,17 @@ export default function NamespaceDetails({ name }) {
   }
 
   return (
-    <>
+    <YamlEditorProvider>
```
Wawrzyn321, line 40: Would it make sense to move it to the index.js?

shared/contexts/YamlEditorContext.js
```diff
@@ -0,0 +1,123 @@
...
+  useEffect(() => {
+    if (!isOpen) setJson(null);
+  }, [isOpen]);
+
+  useEffect(() => {
+    LuigiClient.uxManager().setDirtyStatus(!!changedYaml);
+  }, [changedYaml]);
+
+  function setEditedJson(newJson, onSaveHandler) {
+    onSaveFn.current = onSaveHandler; // MUST BE AN ANSYNC FUNCTION!
```
Wawrzyn321, line 72: I think this requirement won't be needed if we use _await_ with try..catch instead of _then_.

Wawrzyn321 (changes requested):
One more thing - if we open the drawer, introduce some changes and then try to navigate to another view, a dirty state modal pops up as expected, but if we confirm, the nav will be performed, but the drawer stays on.

core-ui/src/components/NamespaceDetails/NamespaceDetails.js
Wawrzyn321: Answer: nope, it wouldn't.

Wawrzyn321 (approved):

PR #2055, opened 2020-11-16
Use Pamela on pod list

parostatkiem:
**Description**

- Remove `Pods` from Core and the navigation
- Add generic Pamela-compatible web connection hooks (get, put, delete, subscribe)
- Add `Pods` view to core-ui
- use Drawer for editing pod YAML
- add `PamelaAPI` env variable

**Related issue(s)**


dariadomagala (changes requested):
Few things, might not be only UI related, but I decided to write everything in just one place:

1. Pod names shouldn't have an underline since it's not a link
2. Before deletion, the action should be confirmed first
3. Showing restart count could be useful maybe?
4. Showing pods error messages would also be helpful
5. When you click the left side luigi navigation on pods, while you're already on the list, you'll get a 500 error with no error message
6. When this error accurs, nothing is shown in the UI, just an empty list of pods
7. Pods statuses in the UI don't match statuses from kubectl

dariadomagala (approved):

PR #2090, opened 2021-01-14
Luigi bump

dariadomagala:
**Description**

- ...
- ...
- ...

**Related issue(s)**


parostatkiem (changes requested):
It works but I get this nasty error when going between _namespace list_ and _namespace overview_ views. 
image

dariadomagala:
@parostatkiem I'm aware of it, and so is the Luigi team. I decided to ignore it since it seems to work perfectly wine regardless of this error. Do we want to wait with the bump until it's fixed in Luigi?

parostatkiem (approved):
No, we don't want to wait.

Issue #2097, opened 2021-01-21
Have an OIDC login flow with PKCE and use this token for K8s access


valentinvieriu:
**Description**

As Sebastian and Mathew, I would like to be able to login with a SSO provider ( OIDC ) to the Console and also to my Kubernetes cluster, so it is more user friendly for me to not have additional logins or tokens
**AC**
 - [ ]  I would be able to login in console using an OIDC certified provider
 - [ ]  I want my token to be as secured as possible, so please use security best practices when comes to SPA
 - [ ]  I would like to use the same OIDC to login in my Kubernetes Cluster
 
**Non Functional Requirements**
- [ ]  The OIDC configuration needs to be configurable at build time and run time
- [ ] We should use the Luigi plugins to handle the flow and choose the best security option available
- [ ] Remove any backend related token impersonation
- [ ] We do not care about the adapting the Graphql layer
- [ ] We would like to get the OIDC configuration via URL parameters and stored in localstorage

**Kyma Goals we are achieving with this**
- Slimmer / faster Kyma
- Kyma console can be a stand alone Open Source project and can work with any K8s api server

Sawthis:
It's part of #2080 Epic.

PR #2100, opened 2021-01-25
OIDC flow

Wawrzyn321:
**Description**

- Update Luigi oidc plugin so we can get that sweet 'none' storage option
- Don't use Dex
- Fetch Console init data with Pamela

List and details views won't work for now, so I've added a mock message on NS details and list.

Todos (in the next iteration):
- pass issuerUrl and clientId
- use in-cluster Pamela address instead of a localhost one

**Related issue(s)**
[Original](https://github.com/kyma-project/console/issues/2097)
[Pamela](https://github.com/kyma-project/pamela/pull/12)

### DIFF
--- a/core/src/luigi-config/navigation/queries.js
+++ b/core/src/luigi-config/navigation/queries.js
@@ -1,63 +1,69 @@
-export const GET_MICROFRONTENDS = `query MicroFrontends($namespace: String!) {
-    microFrontends(namespace: $namespace){
-      name
-      category
-      viewBaseUrl
-      navigationNodes{
-        label
-        navigationPath
-        viewUrl
-        showInNavigation
-        order
-        settings
-        requiredPermissions{
-          verbs
-          resource
-          apiGroup
-        }
-      }
-    }
-  }`;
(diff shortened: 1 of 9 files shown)


core/src/luigi-config/navigation/queries.js
```diff
@@ -1,63 +1,69 @@
...
+    selector: data => ({ backendModules: data.items.map(bM => bM.metadata) })
+  };
+
+  const clusterMicroFrontends = {
+    path: '/apis/ui.kyma-project.io/v1alpha1/clustermicrofrontends',
+    selector: data => ({ clusterMicroFrontends: data.items.map(cMF => ({ ...cMF.spec, ...cMF.metadata })) }),
+  }
+
+  const kymaVersion = {
+    path: '/apis/apps/v1/namespaces/kyma-installer/deployments/kyma-installer',
```
dariadomagala, line 28: Namespace admin for example won't have access to this resource

dariadomagala (approved):
It's fine for now, but keep my comment in mind for the future improvements

PR #2131, opened 2021-02-19
Optimize Nginx and Dockerfiles

parostatkiem:
**Description**
I focused on the _core-ui_ microfrontend because other ones will be removed soon.

#### Few words about the current Nginx performance:
Nginx for every microfrontend will be able to handle 256 connections within each 100ms.

- modify Dockerfiles to use the new `console-nginx` base image for the _serve_ stage and `alpine:3.13` for the _build_ stage 
- optimize `Nginx` configurations to consume noticeably less resources
- `core-ui` build time reduced from ~16min to 13min
- remove `antiheld` from the console codeowners
- update images 

**Related issue(s)**


add-ons/nginx.conf
```diff
@@ -23,17 +25,20 @@ http {
   reset_timedout_connection on;
   server_tokens off;
 
-  server {
 
+  map $status $nonSuccessful {
+      ~^2  0;
+      default 1;
+  }
```
Wawrzyn321, line 32: Just a question - do we also want to treat 3xx statuses as `nonSuccessful`?

parostatkiem: Well, I've never got a 3xx status from the k8s API so I suppose it won't pollute our logs too much.

scripts/makefile-node-common.mk
```diff
@@ -107,3 +107,16 @@ exec:
...
+
+install-root:
+	cd .. && npm ci
+
+install-app:
+	npm ci
+
+# for microfrontends which are not using /common or /components to speed up the build time;
+build-shared:
+	cd ../shared && npm ci && npm run build
```
Wawrzyn321, line 119: Good.

Wawrzyn321 (approved):

PR #2138, opened 2021-03-02
Update lodash 

parostatkiem:
**Description**
https://github.tools.sap/kyma/security-scans/issues/6739
https://github.tools.sap/kyma/security-scans/issues/6743

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/pamela/Makefile
+++ b/pamela/Makefile
@@ -1,0 +1,2 @@
+release: 
+	@echo "This is a dummy job for a component which doesn't exist yet"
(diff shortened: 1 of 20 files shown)


pamela/Makefile
```diff
@@ -0,0 +1,2 @@
+release: 
+	@echo "This is a dummy job for a component which doesn't exist yet"
```
dariadomagala, line 2: why do we add this?

parostatkiem: To let the CI job pass. If the `make release` command fails, the job which is **required** also fails.

dariadomagala (approved):

PR #2143, opened 2021-03-08
Adjust Service Catalog Lists to use Pamela

parostatkiem:
**Description**

- adjust Catalog and Instances views to use k8s API
- drop the `Bound Applications` column from Instances
- drop the whole _ServicePlans_ functionality from Catalog and Instances lists (it should be handled at the details level)

**Related issue(s)**


service-catalog-ui/src/helpers/index.js
```diff
@@ -56,7 +58,8 @@ export function isStringValueEqualToTrue(value) {
 }
 
 export const backendModuleExists = name => {
```
dariadomagala, line 60: if we're not using backendmodules in catalog anyways, please remove it

parostatkiem: I decided to not remove anything which is used in the Details view. This is one of those "things".

dariadomagala (approved):

